Issue #45, opened 2026-02-17
Sponsor relay reports confirmed txs as dropped_replace_by_fee


whoabuddy:
## Summary

The sponsor relay successfully broadcasts sponsored transactions, but misreports their settlement status as `dropped_replace_by_fee`. This causes the upstream x402 server to reject the payment even though it confirmed on-chain, resulting in lost funds with no service delivered.

## Observed Behavior

Sent 4 sponsored sBTC transfer requests via `send_inbox_message` (aibtc.com inbox). The relay returned errors for all 4 attempts, but **3 of the 4 transactions confirmed successfully on-chain**.

### Error responses received

**Attempt 1** — Hiro API rate limit during nonce fetch:
```
SPONSOR_FAILED: Error fetching nonce. Response 429: Too Many Requests.
Attempted to fetch https://api.mainnet.hiro.so/v2/accounts/SP1PMPPVCMVW96FSWFV30KJQ4MNBMZ8MRWR3JWQ7?proof=0
```

**Attempts 2-4** — False RBF report:
```
SETTLEMENT_FAILED: transaction verification failed: unexpected transaction status: dropped_replace_by_fee
```

### On-chain evidence (all confirmed successfully)

| Sender Nonce | Sponsor Nonce | Tx ID | Block | Status |
|-------------|--------------|-------|-------|--------|
| 116 | 148 | `0x5a4c651a43f253cc279e38bc0d46cba116cb712456032eb614068af3df6fc409` | 6623343 | **success** |
| 117 | 149 | `0xeedbfcf6f337871c61e45850ccf98ba60d22926b0093c92d085c51d7ff1a9db4` | 6623352 | **success** |
| 118 | 150 | `0xc009d4b74dff2952b02efaeb5c2692be0f7c8575d1926f31d513f762655d2412` | 6623450 | **success** |

All three are `sbtc-token::transfer` of 100 sats from `SP2GHQRCRMYY4S8PMBR49BEKX144VR437YT42SF3B` to `SPKH9AWG0ENZ87J1X0PBD4HETP22G8W22AFNVF8K`, sponsored by `SP1PMPPVCMVW96FSWFV30KJQ4MNBMZ8MRWR3JWQ7`.

**Result:** 300 sats sBTC spent, 0 messages delivered.

## Root Cause Hypothesis

Two interacting issues:

### 1. Hiro API rate limiting (no API key)

[...]

PR #46, opened 2026-02-17
Fix false dropped_replace_by_fee settlement failures

joelklabo:
## Summary
- add a guarded recovery path for facilitator failures that report `dropped_replace_by_fee`
- derive txid from the sponsored tx (or facilitator txid when present), then query Hiro `extended/v1/tx/{txid}`
- if the tx is confirmed and transfer events match `settle.expectedRecipient`, `settle.minAmount`, token type, and optional sender, return a recovered `confirmed` settlement instead of hard-failing
- preserve normal failure behavior when recovery validation does not pass
- keep facilitator HTTP status in failure responses (instead of forcing `400`)

## Why
Issue #45 reports confirmed sponsored transfers being misclassified as `dropped_replace_by_fee`, causing payment to be rejected upstream even after funds are moved on-chain.

## Validation
- `npm run cf-typegen` (with `.env` from `.env.example`)
- `npm run check`
- manual runtime check against confirmed tx from issue #45 to verify recovered settlement parsing

### DIFF
--- a/src/services/facilitator.ts
+++ b/src/services/facilitator.ts
@@ -8,10 +8,15 @@ import type {
   FacilitatorSettleResponse,
   SettlementResult,
 } from "../types";
+import { hexToBytes } from "@stacks/common";
+import { txidFromBytes } from "@stacks/transactions";
 import { HealthMonitor } from "./health-monitor";
 
 const FACILITATOR_TIMEOUT_MS = 30000; // 30 seconds
 const HEALTH_CHECK_TIMEOUT_MS = 5000; // 5 seconds for health checks
+const HIRO_TIMEOUT_MS = 5000; // 5 seconds for tx lookups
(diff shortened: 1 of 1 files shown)


whoabuddy (commented):
Thanks for this @joelklabo — the RBF recovery approach is solid and the event-based validation pattern is well-structured.

We've been working on a parallel fix in #47 that takes a more structural approach: replacing the facilitator entirely with native settlement in the relay. The relay now verifies payment params locally, broadcasts directly to Stacks, and polls for confirmation with exponential backoff up to 60s. This eliminates the facilitator's 30s timeout as the root cause rather than patching around it.

Since #47 removes the facilitator from the `/relay` hot path, the recovery code here would become unreachable once that merges. Going to close this in favor of #47, but want to call out that your event parsing and transfer validation logic is a pattern we may borrow for future receipt verification improvements.

Appreciate the contribution!

PR #77, opened 2026-02-20
feat: nonce mastery — self-healing gaps, dedup liveness, chaining pressure

### LINKED ISSUE #75: NonceDO must self-heal gaps by filling them, not waiting
## Problem

Nonce gaps in the sponsor wallet cause **complete pipeline stalls**. When nonces are reserved but never broadcast (or broadcast and dropped), the gap blocks all subsequent transactions from confirming. The relay's current 30% failure rate is driven primarily by this.

### What happened (2026-02-20)

Sponsor wallet `SP1PMPPVCMVW96FSWFV30KJQ4MNBMZ8MRWR3JWQ7` had 3 missing nonces (488, 489, 490) with 2 pending txs (486, 487) and head at 491. The entire pipeline was stuck:

```
last_executed_tx_nonce: 485
detected_mempool_nonces: [486, 487]
detected_missing_nonces: [488, 489, 490]   ← GAPS
last_mempool_tx_nonce: 491
possible_next_nonce: 492
```

The resync endpoint reported: `"gaps detected (488,489,490) but nonce will fill naturally"` — they didn't. Manual intervention with `fill-nonce-gaps.ts` was required to unblock the pipeline.

### Impact

- 30% tx failure rate (48/158 in 24h)
- Every new relay tx got nonce ≥492, which can't confirm until gaps are filled
- MCP `send_inbox_message` calls fail with `NONCE_CONFLICT` → agents can't send messages
- Earlier successful txs (e.g. sBTC transfers) were dropped by RBF without replacement

## Root Cause

The NonceDO reconciliation logic (alarm every 5 min) **detects** gaps but does **not fill them**. The current gap recovery path resets the pool counter to the lowest gap, which means the next request gets the gap nonce — but that only works if the gap was caused by a skipped assignment. When the gap is caused by a dropped/never-broadcast tx, the nonce slot is genuinely empty in the mempool and needs an actual transaction to fill it.

### How gaps form

1. **Dropped broadcasts**: Relay sponsors tx with nonce N, broadcast succeeds, but tx is later dropped (RBF, mempool eviction, node restart). Nonce N becomes a gap.
2. **Failed releases**: `releaseNonceDO()` is fire-and-forget via `waitUntil()`. If the worker terminates before the release completes, the nonce stays in `reserved` forever — never returned to `available`, never broadcast.
3. **Multi-wallet race**: With wallet rotation, concurrent requests to different wallets can create ordering gaps if one wallet's tx is dropped but others succeed.

## Proposed Solution: Active Gap Filling

The NonceDO alarm (or a dedicated self-heal pass) should **actively fill gaps** by broadcasting minimal STX transfers when gaps are detected.

### Design

```
alarm() / reconcile():
  1. Fetch nonce state from Hiro for each wallet
  2. If detected_missing_nonces is non-empty:
     a. For each gap nonce, broadcast: makeSTXTokenTransfer({
          recipient: FLUSH_RECIPIENT,  // client testing wallet
          amount: 1n,                  // 1 µSTX
          nonce: gapNonce,
          fee: 30000n,                 // high enough for RBF replacement
        })
     b. Log the fill attempt and result
     c. If broadcast succeeds → gap filled, pipeline unblocked
     d. If ConflictingNonceInMempool → gap already occupied, skip
  3. After filling, resync pool state
```

### Key requirements

1. **Automatic**: No human intervention. The alarm detects and fills gaps within one cycle (5 min).
2. **Scale with N wallets**: Each wallet in the rotation pool gets independent gap detection and filling. `SPONSOR_WALLET_COUNT` determines the iteration range.
3. **Safe**: Sending 1 µSTX to a known address is the minimal possible cost. Fee of 30,000 µSTX (0.03 STX) per fill is acceptable — gaps are rare and the cost of a stuck pipeline is much higher.
4. **Idempotent**: If a gap is already occupied (ConflictingNonceInMempool), skip it. If the fill tx is later dropped, the next alarm cycle will re-fill.
5. **Observable**: Log every gap fill with wallet index, nonce, txid, and reason. Increment a `gapsFilled` counter in stats.

### Implementation details

- **Flush recipient**: Use env var `FLUSH_RECIPIENT` (default: `SPEB8Z3TAY2130B8M5THXZEQQ4D6S3RMYT37WTAC`). Cannot send to self.
- **Fee**: Use the current clamped fee from FeeService, with a minimum floor of 30,000 µSTX for RBF headroom.
- **Signing**: NonceDO needs access to the sponsor private key (or delegate to SponsorService). Currently the DO only tracks nonces — it doesn't sign. Options:
  - A) Pass the sponsor key to NonceDO via env (simplest, key already in secrets)
  - B) NonceDO calls back to the worker to request a signed fill tx (more isolated)
  - C) New `POST /nonce/fill` endpoint that the alarm triggers (keeps signing in the worker)
- **Wallet rotation**: The alarm already iterates over all wallets (`for wi = 0..MAX_WALLET_COUNT`). Gap filling should use the same per-wallet iteration, deriving the correct key for each wallet index.
- **Backoff**: If Hiro API is down, skip the fill (same as current alarm behavior). Don't retry fills within the same alarm cycle.

### Changes needed

1. `src/durable-objects/nonce-do.ts` — `reconcileNonceForWallet()`: When gaps are detected, fill them instead of logging "will fill naturally"
2. `src/services/sponsor.ts` — Add `fillNonceGap(walletIndex, nonce)` method that builds and broadcasts a minimal STX transfer
3. `wrangler.jsonc` — Add `FLUSH_RECIPIENT` env var
4. Stats: Add `gapsFilled` counter to StatsService
5. Tests: Verify gap filling works across multiple wallets

### What we have today (manual)

- `scripts/fill-nonce-gaps.ts` — Works perfectly for manual recovery but requires human detection + `.env` access
- `POST /nonce/reset` with `action: resync` — Detects gaps but doesn't fill them
- 5-minute alarm — Same detection without action

The gap between detection and action is the bug.

## References

- PR #64 — NonceDO introduction
- PR #67 — Gap detection alarm
- PR #70 — Nonce fail-fast
- PR #74 — Multi-wallet rotation + nonce pool
- `scripts/fill-nonce-gaps.ts` — Manual recovery tool (proven to work)

### LINKED ISSUE #76: Dedup cache poisons retries after NONCE_CONFLICT
## Problem

When a transaction fails with `NONCE_CONFLICT`, the relay's dedup cache stores the failed result keyed on the original agent transaction hex. All subsequent retries of the same agent payload hit the dedup cache and receive the **stale sponsoredTx** from the first failed attempt — including a dead txid that was never confirmed and may have been dropped.

This makes `NONCE_CONFLICT` errors **unrecoverable** without waiting for the 5-minute dedup TTL to expire.

### Observed behavior

1. Agent sends sBTC transfer via `/relay` → relay sponsors with nonce N → broadcast succeeds → nonce conflict → returns `NONCE_CONFLICT` with txid `f8d1e3...`
2. Relay stores dedup entry: `sha256(body.transaction) → { txid: f8d1e3, sponsoredTx: <nonce-N-version>, ... }`
3. Agent retries same payload → **dedup hit** → returns cached `{ txid: f8d1e3, sponsoredTx: <stale> }`
4. txid `f8d1e3...` is dead — rejected by node with `BadNonce`, never entered mempool
5. Every retry for 5 minutes returns the same dead txid

### Root cause

The dedup key is `sha256(body.transaction)` — the **original unsigned agent transaction**. The comment at `src/services/settlement.ts:729-731` says:

> Note: dedup keys are SHA-256 of the fully-sponsored tx hex. A resubmitted transaction after a NONCE_CONFLICT will have a new sponsor nonce and therefore a different hex — it will not match this cache entry.

**This comment is incorrect.** The actual dedup is on the unsigned agent tx, not the sponsored version. Since the agent's payload doesn't change between retries (same sender, recipient, amount), every retry matches the cached entry.

### Impact

- MCP `send_inbox_message` becomes permanently stuck after one NONCE_CONFLICT
- Agents must wait 5 minutes for dedup TTL before retrying
- The `retryable: true, retryAfter: 1` in the NONCE_CONFLICT response is misleading — retrying after 1 second hits the same dedup cache

### Proposed fix

**Option A (recommended): Don't cache NONCE_CONFLICT results in dedup**

Only store dedup entries for transactions that were successfully broadcast (have a valid txid in mempool or confirmed). NONCE_CONFLICT means the sponsored version was rejected — caching it prevents recovery.

```typescript
// In settlement flow, after broadcast:
if (broadcastResult.nonceConflict) {
  // Do NOT store dedup — let next retry get a fresh sponsor nonce
  releaseNonceDO(...);
  return err({ code: "NONCE_CONFLICT", ... });
}

// Only cache on successful broadcast:
await storeDedup(txHash, { txid, sponsoredTx, ... });
```

**Option B: Key dedup on fully-sponsored tx hex (match the comment)**

Change the dedup key from `sha256(body.transaction)` to `sha256(sponsoredTxHex)`. Each sponsoring attempt produces a different hex (different nonce), so retries won't match. This preserves dedup for true duplicates (same sponsored tx submitted twice).

**Option C: Add dedup bypass for retries**

Accept a `retry: true` field in the request body that tells the relay to skip the dedup check. The MCP tool would set this on subsequent attempts.

### Related

- #75 — NonceDO should self-heal gaps (root cause of the NONCE_CONFLICT)
- The combination of #75 + this issue makes nonce problems cascade: gaps cause conflicts, conflicts poison the cache, cache prevents recovery

whoabuddy:
## Summary

- **Phase 1 — NonceDO active gap filling (#75):** The 5-minute alarm now actively fills detected nonce gaps by broadcasting 1 uSTX transfers to `FLUSH_RECIPIENT` instead of logging "will fill naturally". Per-wallet key derivation, `gapsFilled` counter, and `FLUSH_RECIPIENT` env var added.
- **Phase 2 — Fix dedup cache poisoning (#76):** `checkDedup()` now verifies pending txids are still alive via Hiro API before returning cached results. Dead txids trigger cache invalidation (KV delete + cache miss). Fixed misleading comment about dedup key scheme.
- **Phase 3 — Reduce chaining pressure:** Stale reservation cleanup (>10 min orphans reclaimed by alarm), `CHAINING_LIMIT` reduced from 25 to 20 for safety buffer, `ChainingLimitError` carries mempool depth + estimated drain time in 429 responses.
- **Config:** `SPONSOR_WALLET_COUNT` increased to 5 (all funded), testnet `FLUSH_RECIPIENT` corrected to `ST` prefix.
- **Refactor:** Extracted helpers (`insertSorted`, `isTxStatusTerminal`, `truncateHash`, `NonceDOErrorBody`), named constants, normalized log style.

Closes #75, closes #76

## Test plan

- [ ] `npm run check` — type-check passes
- [ ] `npm run deploy:dry-run` — build succeeds
- [ ] Deploy to staging, create a nonce gap, verify alarm fills it within 5 minutes
- [ ] Verify dedup cache returns fresh results after a pending tx dies
- [ ] Confirm stale reservations are reclaimed after 10 minutes
- [ ] Verify 5-wallet rotation under load (chaining limit = 20 × 5 = 100 concurrent)

### DIFF
--- a/src/durable-objects/nonce-do.ts
+++ b/src/durable-objects/nonce-do.ts
@@ -144,8 +144,11 @@ const VALID_MNEMONIC_LENGTHS = [12, 24];
 const GAP_FILL_AMOUNT = 1n;
 /** Gap-fill fee: 30,000 uSTX (high enough for RBF priority) */
 const GAP_FILL_FEE = 30_000n;
-/** Default recipient for gap-fill self-transfers */
-const DEFAULT_FLUSH_RECIPIENT = "SPEB8Z3TAY2130B8M5THXZEQQ4D6S3RMYT37WTAC";
+/** Default recipient for gap-fill self-transfers, per network */
+const DEFAULT_FLUSH_RECIPIENT_MAINNET = "SPEB8Z3TAY2130B8M5THXZEQQ4D6S3RMYT37WTAC";
+const DEFAULT_FLUSH_RECIPIENT_TESTNET = "STEB8Z3TAY2130B8M5THXZEQQ4D6S3RMYRENN2KB";
+/** Maximum number of gap-fill broadcasts per alarm cycle per wallet */
+const MAX_GAP_FILLS_PER_ALARM = 5;
 
 // Legacy single-wallet pool key (migrated to pool:0 on first access)
 const POOL_KEY = "pool";
@@ -415,7 +418,10 @@ export class NonceDO {
     privateKey: string
   ): Promise<string | null> {
     const network = this.env.STACKS_NETWORK ?? "testnet";
-    const recipient = this.env.FLUSH_RECIPIENT ?? DEFAULT_FLUSH_RECIPIENT;
+    const defaultRecipient = network === "mainnet"
+      ? DEFAULT_FLUSH_RECIPIENT_MAINNET
+      : DEFAULT_FLUSH_RECIPIENT_TESTNET;
+    const recipient = this.env.FLUSH_RECIPIENT ?? defaultRecipient;
     try {
       const tx = await makeSTXTokenTransfer({
         recipient,
@@ -968,11 +974,13 @@ export class NonceDO {
 
       this.setStateValue(STATE_KEYS.lastGapDetected, Date.now());
 
-      // Actively fill gaps: derive key and broadcast 1 uSTX transfers for each gap nonce
+      // Actively fill gaps: derive key and broadcast 1 uSTX transfers for each gap nonce.
+      // Cap per-alarm fills to avoid exceeding Cloudflare alarm execution time limits.
       const privateKey = await this.derivePrivateKeyForWallet(walletIndex);
       const filled: number[] = [];
+      const gapsToFill = sortedGaps.slice(0, MAX_GAP_FILLS_PER_ALARM);
       if (privateKey) {
-        for (const gapNonce of sortedGaps) {
+        for (const gapNonce of gapsToFill) {
           const txid = await this.fillGapNonce(walletIndex, gapNonce, privateKey);
           if (txid) {
             console.log(
--- a/src/services/settlement.ts
+++ b/src/services/settlement.ts
@@ -40,6 +40,8 @@ const MAX_POLL_DELAY_MS = 8_000;
 // KV dedup configuration
 const DEDUP_TTL_SECONDS = 300;
 const DEDUP_KEY_PREFIX = "dedup:";
+/** Only verify liveness of pending dedup entries older than this (ms) */
+const DEDUP_LIVENESS_AGE_MS = 60_000;
 
 /** Shape of Hiro GET /extended/v1/tx/{txid} response (subset) */
 interface HiroTxResponse {
@@ -804,10 +806,11 @@ export class SettlementService {
 
       const result = JSON.parse(cached) as DedupResult;
 
-      // For "pending" entries, verify the txid is still alive before returning.
-      // If the transaction was dropped or replaced, the cached entry is stale and
-      // must be invalidated so the caller can submit a fresh broadcast.
-      if (result.status === "pending") {
+      // For "pending" entries older than DEDUP_LIVENESS_AGE_MS, verify the txid is
+      // still alive. Fresh entries are trusted to avoid adding Hiro API latency to
+      // the hot path on every dedup hit.
+      const entryAge = result.recordedAt ? Date.now() - result.recordedAt : Infinity;
+      if (result.status === "pending" && entryAge > DEDUP_LIVENESS_AGE_MS) {
         const alive = await this.verifyTxidAlive(result.txid);
         if (!alive) {
           this.logger.warn("Dedup stale: pending txid is dead, invalidating cache entry", {
@@ -848,9 +851,11 @@ export class SettlementService {
     try {
       const txHash = await this.computeTxHash(sponsoredTxHex);
       const key = `${DEDUP_KEY_PREFIX}${txHash}`;
-      await this.env.RELAY_KV.put(key, JSON.stringify(result), {
-        expirationTtl: DEDUP_TTL_SECONDS,
-      });
+      await this.env.RELAY_KV.put(
+        key,
+        JSON.stringify({ ...result, recordedAt: Date.now() }),
+        { expirationTtl: DEDUP_TTL_SECONDS }
+      );
 
       this.logger.debug("Dedup result recorded in KV", {
         txHash: this.truncateHash(txHash),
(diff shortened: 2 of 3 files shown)


whoabuddy:
@pbtc21 — thanks for the field report, that timeline is invaluable for validating the fix.

The three root causes you hit map directly to the phases in this PR:
- **TooMuchChaining** → Phase 3 (chaining limit reduced to 20, stale reservations cleaned, 5 wallets now active)
- **ConflictingNonceInMempool from dedup** → Phase 2 (liveness check invalidates dead txids)
- **BadNonce during gap fill catch-up** → Phase 1 (alarm actively fills gaps now)

Your feedback on dedup caching latency was incorporated — liveness checks now only fire on entries > 60s old per cedarxyz's review.

Let us know once the queued messages go through. If Rapid Vera, Mystic Core, or Mighty Scorpion are still stuck after the next alarm cycle (~5 min), ping here and we'll dig into the specific nonces.

PR #106, opened 2026-02-22
fix(settlement): pass maxTimeoutSeconds through to broadcastAndConfirm

### LINKED ISSUE #105: fix: cascading timeout causes 500 empty body on sBTC settlement
## Problem

Services calling the relay's `/settle` endpoint get 500 empty-body responses during sBTC payments when the Hiro API is slow. The cascade:

1. **Hiro API** responds slowly (intermittent, 10-20s+ latency spikes)
2. **x402-sponsor-relay** `/settle` → `broadcastAndConfirm()` polls for up to **60s** (hardcoded `MAX_POLL_TIME_MS`)
3. **Calling service** (e.g., aibtc.com inbox) uses `x402-stacks` library with **30s default timeout** → fetch times out before relay responds → unhandled exception → 500 empty body

The relay itself handles the timeout gracefully (returns `{ status: "pending" }` after 60s), but callers time out before that response arrives.

## Evidence

- **27+ hours** of 500 empty-body responses observed (Feb 21-22, 2026)
- 402 challenge handshake works every time — payment POST fails
- Transactions are well-formed (valid nonce, hex, sufficient sBTC balance)
- Worker-logs show error spikes correlating with Hiro API slowness:
  - x402-relay: 136 errors on Feb 18 spike, 2 by Feb 22
  - x402-api-host: 49 errors on Feb 18 spike, 5 by Feb 22
- x402-api applies a patch (`x402-stacks+2.0.1.patch`) bumping timeouts to 120s; other callers don't have this patch

## Root Cause

Mismatch between relay poll time (60s max) and `x402-stacks` library default timeouts (30s settle, 15s verify). The relay's `/settle` can take up to 60s+ for sBTC contract calls, but callers using the library's defaults time out at 30s.

## Proposed Fix

1. Accept `maxTimeoutSeconds` from `paymentRequirements` in `/settle` and cap `broadcastAndConfirm()` polling accordingly
2. Default remains 60s, but callers requesting shorter timeouts (e.g., 30s) get an early `{ status: "pending" }` response instead of the caller timing out

This way the relay responds within the caller's budget and the caller gets a valid response (pending) instead of a crash.

## Related

- PR #103 improved Hiro API timeouts (fee, nonce, broadcast) but didn't address the poll-time/caller-timeout mismatch
- `x402-api` works around this with a 120s timeout patch on x402-stacks

whoabuddy:
## Summary

- Accept `maxTimeoutSeconds` from `paymentRequirements` in `/settle` and from `settle` options in `/relay`
- Pass it through to `broadcastAndConfirm()` as `maxPollTimeMs` so the relay responds with `{ status: "pending" }` within the caller's timeout budget
- `broadcastAndConfirm` now accepts optional `maxPollTimeMs` parameter, capped at the existing 60s `MAX_POLL_TIME_MS`

This prevents the cascading timeout where callers (using x402-stacks library with 30s default) time out before the relay's 60s poll completes, producing 500 empty-body errors.

## Test plan

- [ ] Deploy to staging and test `/settle` with `maxTimeoutSeconds: 30` in `paymentRequirements`
- [ ] Verify relay returns `{ status: "pending" }` instead of caller timing out when Hiro is slow
- [ ] Test `/relay` with `maxTimeoutSeconds` in settle options
- [ ] Confirm default behavior (no `maxTimeoutSeconds`) still polls up to 60s

Closes #105

### DIFF
--- a/src/endpoints/relay.ts
+++ b/src/endpoints/relay.ts
@@ -356,9 +356,11 @@ export class Relay extends BaseEndpoint {
       // Step D — Broadcast and poll for confirmation.
       // Cap poll time to caller's maxTimeoutSeconds so the relay responds
       // before the caller's own timeout fires (avoids 500 empty-body errors).
+      // Reserve 5s for broadcast overhead, sponsoring, and response serialization.
+      const RELAY_OVERHEAD_MS = 5_000;
       const maxPollTimeMs =
         body.settle.maxTimeoutSeconds != null && body.settle.maxTimeoutSeconds > 0
-          ? body.settle.maxTimeoutSeconds * 1000
+          ? Math.max(body.settle.maxTimeoutSeconds * 1000 - RELAY_OVERHEAD_MS, 1_000)
           : undefined;
       const broadcastResult = await settlementService.broadcastAndConfirm(
         verifyResult.data.transaction,
--- a/src/endpoints/settle.ts
+++ b/src/endpoints/settle.ts
@@ -162,8 +162,10 @@ export class Settle extends BaseEndpoint {
       // Broadcast and poll for confirmation.
       // Cap poll time to caller's maxTimeoutSeconds (from paymentRequirements)
       // so the relay responds before the caller's own timeout fires.
+      // Reserve 5s for broadcast overhead and response serialization.
+      const SETTLE_OVERHEAD_MS = 5_000;
       const maxPollTimeMs = validation.data.maxTimeoutSeconds != null
-        ? validation.data.maxTimeoutSeconds * 1000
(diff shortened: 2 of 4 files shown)


whoabuddy:
All 4 Copilot review comments addressed in f062e13:

1. **Type mismatch**: Made `maxTimeoutSeconds` optional in `X402PaymentRequirementsV2` type
2. **Broadcast overhead (settle)**: Reserved 5s buffer when computing `maxPollTimeMs` 
3. **Broadcast overhead (relay)**: Same 5s buffer applied
4. **Polling loop budget**: Capped sleep delay and per-request timeout by remaining time; added fallback return for break-path exits

Type-check passes clean.

PR #136, opened 2026-03-02
fix(btc-verify): remove varint prepend from bip322TaggedHash

### LINKED ISSUE #135: Non standard hash still for btc signing?
I finally got back to verification and heartbeat today for my agent but it still wasnt working after the updates ... after a ton of analysis, claude gave me this ... my heartbeat script is now finally running consistently, not that you really care about that per se, but it implies that the `non-standard` fix that was applied in my heartbeat is "correct".

ref #69 

```
---
  Root cause: The AIBTC relay server's bip322TaggedHash function has a bug — it
  prepends varint(message.length) before the message bytes when computing the
  BIP322 tagged hash:

  Server (buggy):  SHA256(tagHash || tagHash || varint(41) || message_bytes)
  BIP322 spec:     SHA256(tagHash || tagHash || message_bytes)

  This produces a different toSpend txid → different BIP143 sighash → signatures
  from spec-compliant libraries like bip322-js will always fail verification.

  Fix: The script now uses a custom aibtcSign() function (backed by
  @scure/btc-signer and @stacks/encryption, the same libraries the server uses)
  that replicates the server's non-standard hash. The bip322-js import is no
  longer needed.

  Bonus: removed bip322-js, ecpair, and bitcoinjs-lib from the signing path
  entirely — key derivation still uses them, but signing now goes through
  @scure/btc-signer directly.

  You may also want to file a bug with AIBTC pointing at bip322TaggedHash in
  x402-sponsor-relay/src/services/btc-verify.ts — the varint prepend isn't in the
   BIP322 spec and will break any third-party client that implements the standard
   correctly.
```

arc0btc:
## Summary

Fixes #135 — removes the incorrect `varint(msg.length)` prepend from `bip322TaggedHash` in `src/services/btc-verify.ts`.

### Root Cause

BIP-322 uses BIP-340 tagged hashes, which have the form:

```
SHA256(SHA256(tag) || SHA256(tag) || msg)
```

The varint length prefix is a Bitcoin message format convention (BIP-137) — it does **not** belong in a BIP-340 tagged hash. Including it produces a different `toSpend` txid → different BIP-143/BIP-341 sighash → signatures from any spec-compliant BIP-322 library always fail verification.

### The Fix

```diff
 function bip322TaggedHash(message: string): Uint8Array {
   const tagBytes = new TextEncoder().encode("BIP0322-signed-message");
   const tagHash = hashSha256Sync(tagBytes);
   const msgBytes = new TextEncoder().encode(message);
-  const varint = encodeVarInt(msgBytes.length);
-  const msgPart = concatBytes(varint, msgBytes);
-  return hashSha256Sync(concatBytes(tagHash, tagHash, msgPart));
+  return hashSha256Sync(concatBytes(tagHash, tagHash, msgBytes));
 }
```

`encodeVarInt` is still used in other parts of the file (BIP-143 transaction encoding) and is not removed.

### Impact

- **BIP-137 verification** (P2PKH `1...`, P2SH `3...`) is **unaffected** — uses `formatBitcoinMessage`, not `bip322TaggedHash`
- **BIP-322 verification** (P2WPKH `bc1q...`, P2TR `bc1p...`) will now verify signatures from spec-compliant clients correctly
- **Coordinated fix** — same bug exists in 3 other repos (see issue #135 for links): aibtc-mcp-server, landing-page, skills

### DIFF
--- a/src/services/btc-verify.ts
+++ b/src/services/btc-verify.ts
@@ -81,7 +81,7 @@ function detectBtcNetwork(address: string): BTC_NETWORK {
  * Result of BTC signature verification
  */
 export type BtcVerifyResult =
-  | { valid: true; path: "registration" | "self-service"; timestamp?: string }
+  | { valid: true; path: "registration" | "self-service"; timestamp?: string; deprecationNotice?: string }
   | { valid: false; error: string; code: BtcVerifyErrorCode };
 
 /**
@@ -305,12 +305,12 @@ function verifyBip137(address: string, message: string, signatureBase64: string)
 // ---------------------------------------------------------------------------
 
 /**
- * BIP-322 tagged hash: SHA256(SHA256(tag) || SHA256(tag) || msg)
+ * Spec-compliant BIP-322 tagged hash: SHA256(SHA256(tag) || SHA256(tag) || msg)
  * where tag = "BIP0322-signed-message"
  *
- * Note: BIP-322 uses BIP-340 tagged hashes, which do NOT include a varint
- * length prefix. The varint is a Bitcoin message format thing (BIP-137) and
- * does not belong here. Fixes: https://github.com/aibtcdev/x402-sponsor-relay/issues/135
+ * BIP-322 uses BIP-340 tagged hashes, which do NOT include a varint length prefix.
+ * The varint is a Bitcoin message format convention (BIP-137) and does not belong here.
+ * Fixes: https://github.com/aibtcdev/x402-sponsor-relay/issues/135
  */
 function bip322TaggedHash(message: string): Uint8Array {
   const tagBytes = new TextEncoder().encode("BIP0322-signed-message");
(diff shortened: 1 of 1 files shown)


whoabuddy (changes requested):
## Backward compatibility needed

This fix is correct per BIP-322 spec, but it will break agents still using the old (non-standard) signing method. We need to support both formats during a transition period.

### Requested changes

**Keep both hash functions:**
```typescript
/** Spec-compliant BIP-322 tagged hash (BIP-340) */
function bip322TaggedHash(message: string): Uint8Array {
  const tagBytes = new TextEncoder().encode("BIP0322-signed-message");
  const tagHash = hashSha256Sync(tagBytes);
  const msgBytes = new TextEncoder().encode(message);
  return hashSha256Sync(concatBytes(tagHash, tagHash, msgBytes));
}

/** @deprecated Legacy tagged hash with varint prepend — non-standard, remove after transition */
function bip322TaggedHashLegacy(message: string): Uint8Array {
  const tagBytes = new TextEncoder().encode("BIP0322-signed-message");
  const tagHash = hashSha256Sync(tagBytes);
  const msgBytes = new TextEncoder().encode(message);
  const varint = encodeVarInt(msgBytes.length);
  const msgPart = concatBytes(varint, msgBytes);
  return hashSha256Sync(concatBytes(tagHash, tagHash, msgPart));
}
```

**In `bip322BuildToSpendTxId`, try spec-compliant first, fall back to legacy:**
- Try building `toSpend` with the correct hash → verify signature
- If that fails, retry with the legacy hash → verify signature
- If the legacy path succeeds, log a deprecation warning and surface a hint in the response (e.g., `"deprecationNotice": "BIP-322 signature uses non-standard tagged hash. Update to spec-compliant signing — see aibtcdev/skills or install latest @aibtc/mcp-server."`)

**Why:** Agents may have the old MCP server or custom signing code baked in. Cutting them off with no transition path is worse than a deprecation hint.

arc0btc (commented):
Implemented exactly as requested in 8e1cc7f.

**What changed:**

Both hash functions kept:
- `bip322TaggedHash` — spec-compliant (no varint)
- `bip322TaggedHashLegacy` — `@deprecated`, varint prepend preserved for transition

`bip322BuildToSpendTxId` refactored to accept a pre-computed `msgHash: Uint8Array` instead of `message: string`, so the two verifier paths can inject different hashes cleanly.

Each address type has a Core inner function (`bip322VerifyP2WPKHCore`, `bip322VerifyP2TRCore`) that takes the hash directly. The outer wrappers try spec-compliant first, fall back to legacy:

```typescript
function bip322VerifyP2WPKH(message, sig, address): { verified, usedLegacy? } {
  if (bip322VerifyP2WPKHCore(sig, address, bip322TaggedHash(message))) return { verified: true };
  if (bip322VerifyP2WPKHCore(sig, address, bip322TaggedHashLegacy(message))) return { verified: true, usedLegacy: true };
  return { verified: false };
}
```

`BtcVerifyResult` updated with `deprecationNotice?: string` on the valid variant. When legacy path succeeds, the response includes:
```
"deprecationNotice": "BIP-322 signature uses non-standard tagged hash. Update to spec-compliant signing — see aibtcdev/skills or install latest @aibtc/mcp-server."
```

A `logger.warn` fires on legacy path too for server-side observability.

TypeScript compiles clean (only pre-existing `worker-configuration.d.ts` error from wrangler types, unrelated to this change).

whoabuddy (approved):
Looks good — spec-compliant hash with legacy fallback and deprecation notice. CI passing.

PR #158, opened 2026-03-12
feat: production hardening — stuck-tx RBF and broadcast failover

whoabuddy:
### Root cause fix (b7271c5) — nonce conflict infinite loop

**The bug:** When a broadcast is rejected with `ConflictingNonceInMempool`, both `sponsor.ts` and `relay.ts` released the nonce with no txid. The DO's `releaseNonce()` saw no txid record and returned the nonce to `available[]`. But the nonce slot was still occupied in the mempool. Next request got the same nonce, broadcast, conflict, release, repeat — infinite loop.

**The fix:** Nonce conflict rejections now pass a synthetic txid marker (`conflict:quarantine:<nonce>`) so `releaseNonce()` moves the nonce to `spent[]` (permanent quarantine). The pool naturally advances past it on the next alarm cycle. Non-conflict failures (5xx, timeouts) still return nonces to `available[]` since they were never broadcast.

### Supporting improvements

- **Stuck-transaction RBF** (56528b7): Detects transactions stuck in mempool >15min via Hiro API and broadcasts replace-by-fee self-transfers at 90K uSTX to clear the slot. Capped at 3 attempts per nonce.
- **Multi-node broadcast failover** (14dca77): Infrastructure for `BROADCAST_NODE_URLS` env var. Optional/future-proofing — only useful if you run your own Stacks node.
- **Fee-fetch fallback** (ca55e1d): When Hiro is unavailable, fees degrade to hardcoded defaults instead of blocking the broadcast flow.
- **Polling resilience** (ca55e1d): `polling_degraded` warn after 3 consecutive poll failures.
- **Nonce conflict retry backoff** (f9573c3): Increased backoff and added /health guidance in error responses.

## Production Evidence

Logs from `logs.aibtc.com` (Mar 11-12, 2026) showed the infinite loop:
```
nonce_released: wallet 1, nonce 500, consumed: false, txid: null  (13:36)
nonce_released: wallet 1, nonce 500, consumed: false, txid: null  (17:28)
nonce_released: wallet 1, nonce 500, consumed: false, txid: null  (18:36)
```
Same nonce 500 released back to available[] 3+ times — each time re-assigned and re-rejected with `ConflictingNonceInMempool`. The pool showed `poolReserved: 0` because the nonce was always released between attempts.

## Deploy notes

- No new secrets required — `BROADCAST_NODE_URLS` is optional
- After deploy, nonce conflicts will quarantine instead of loop
- Monitor for `nonce_quarantined` logs with reason `txid_recorded_on_failed_release`
- `nonce_reconcile_stale` warnings should stop once quarantined nonces are skipped

## Test plan

- [x] `npm run check` (TypeScript) passes
- [ ] Deploy to staging and trigger a nonce conflict — verify nonce goes to spent[] not available[]
- [ ] Confirm `nonce_reconcile_stale` loop breaks after deploy
- [ ] Monitor `stuckTxRbfAttempted` stats for the supplementary RBF mechanism

### DIFF
--- a/src/durable-objects/nonce-do.ts
+++ b/src/durable-objects/nonce-do.ts
@@ -186,8 +186,8 @@ interface NonceStatsResponse {
   chainingLimit: number;
   /** Per-wallet pool state (multi-wallet rotation) */
   wallets: WalletPoolStats[];
-  /** Total RBF broadcast attempts for stuck mempool transactions */
-  stuckTxRbfAttempted: number;
+  /** Total RBF replacements successfully broadcast for stuck mempool transactions */
+  stuckTxRbfBroadcast: number;
   /** Total nonces successfully unstuck via RBF (confirmed after replacement) */
   stuckTxRbfConfirmed: number;
 }
@@ -275,7 +275,7 @@ const STATE_KEYS = {
   gapsFilled: "gaps_filled",
   lastHiroSync: "last_hiro_sync",
   lastGapDetected: "last_gap_detected",
-  stuckTxRbfAttempted: "stuck_tx_rbf_attempted",
+  stuckTxRbfBroadcast: "stuck_tx_rbf_attempted",
   stuckTxRbfConfirmed: "stuck_tx_rbf_confirmed",
 } as const;
 
@@ -469,9 +469,9 @@ export class NonceDO {
     this.setStateValue(STATE_KEYS.gapsFilled, gapsFilled);
   }
 
-  private incrementStuckTxRbfAttempted(): void {
-    const count = this.getStoredCount(STATE_KEYS.stuckTxRbfAttempted) + 1;
-    this.setStateValue(STATE_KEYS.stuckTxRbfAttempted, count);
+  private incrementStuckTxRbfBroadcast(): void {
+    const count = this.getStoredCount(STATE_KEYS.stuckTxRbfBroadcast) + 1;
+    this.setStateValue(STATE_KEYS.stuckTxRbfBroadcast, count);
   }
 
   private incrementStuckTxRbfConfirmed(): void {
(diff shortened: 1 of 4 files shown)


arc0btc:
Reviewing Copilot's comments since my earlier approval. A few worth flagging:

**Agree — [suggestion] `/health` endpoint guidance is wrong** (relay.ts + sponsor.ts)

Copilot is correct: `/health` only returns `{status, network, version}` — clients won't find nonce pool state there. The error message `"back off and retry after checking GET /health for nonce pool status"` is misleading. Should reference `/nonce/stats` (or whatever exposes `circuitBreakerOpen` / pool state). Same issue in both the relay.ts and sponsor.ts changes. Worth fixing before merge — bad endpoint guidance in error messages gets cargo-culted into clients.

**Disagree — NaN date handling** (nonce-do.ts)

Copilot says a bad `receipt_time_iso` would cause "NaN < threshold is false, potentially triggering immediate RBF." The direction is reversed: `NaN < STUCK_TX_AGE_MS` evaluates to `false`, so a bad timestamp would **skip** RBF — the stuck nonce silently goes undetected. Not immediate-fire, but silent miss. In practice the Hiro mempool API always returns valid ISO strings, so this is low risk, but `Number.isFinite(ageMs) && ageMs > STUCK_TX_AGE_MS` is the right guard.

**Agree — nit** per-alarm RBF cap

The current loop can fire RBF for every reserved nonce in one alarm cycle. During a bad outage that could be several. A cap (1–2 per alarm, lowest nonce first) keeps alarm runtime predictable. Minor for now since `MAX_RBF_ATTEMPTS = 3` limits total exposure, but worth a follow-up.

**On the RBF → single-node path** (my earlier `[question]`)

I noted that `broadcastRbfForNonce` uses `@stacks/transactions` directly (Hiro only) rather than the new multi-node path. Given that RBF is triggered from DO alarm reconciliation, not from a request path, this seems intentional — but worth a note in the code if so, since it means RBF is still Hiro-dependent even when `BROADCAST_NODE_URLS` is configured.

CI is green (both production and staging builds pass). The core nonce-conflict quarantine fix is solid and urgently needed. The `/health` endpoint guidance is the one thing I'd fix before shipping to production.

PR #162, opened 2026-03-12
feat: support self-pay (non-sponsored) transaction settlement via X-Settlement header

### LINKED ISSUE #128: Support non-sponsored (self-pay) transaction settlement
## Context

Moved from aibtcdev/aibtc-mcp-server#179 — this is a relay/facilitator concern, not an MCP client concern.

## Problem

Currently the relay only supports sponsored transactions. Users who want to pay their own transaction fees (non-sponsored settlement) have no path to do so.

## Expected Behavior

The relay should support a mode where the original signer pays their own fees, without requiring a sponsor.

## References

- Original issue: https://github.com/aibtcdev/aibtc-mcp-server/issues/179

secret-mars:
## Summary

Closes #128

Adds `X-Settlement: self-pay` header support to the `/relay` endpoint. When this header is present, the relay skips sponsoring entirely — the caller provides a fully-signed standard (non-sponsored) transaction and covers their own network fees.

- **`src/endpoints/relay.ts`**: Detects the `X-Settlement: self-pay` header after settle options validation and delegates to a new private `handleSelfPay()` method. The existing sponsored path is completely unchanged — no behaviour change for callers that omit the header.
- **`src/services/sponsor.ts`**: Adds `validateNonSponsoredTransaction()` which deserialises the tx and explicitly rejects it if `authType === AuthType.Sponsored` (caller error — a sponsored tx makes no sense for self-pay).

The self-pay path mirrors the sponsored flow step-by-step:
1. Validate + deserialize tx (standard auth only)
2. Rate limit by sender (same policy)
3. Dedup check (keyed on original tx for idempotent retries)
4. `SettlementService.verifyPaymentParams()` — auth-agnostic, no changes needed
5. `SettlementService.broadcastAndConfirm()` — auth-agnostic, no changes needed
6. Stats recording, receipt storage, dedup entry
7. Response — same shape as sponsored path; `sponsoredTx` is omitted (undefined) since no sponsor signature was applied

## Test plan

- [ ] `POST /relay` without `X-Settlement` header continues to sponsor and settle as before (backward compat)
- [ ] `POST /relay` with `X-Settlement: self-pay` and a standard auth tx: verifies payment params and broadcasts directly, response has no `sponsoredTx` field
- [ ] `POST /relay` with `X-Settlement: self-pay` and a sponsored tx: rejected with `400 INVALID_TRANSACTION` — "Transaction must not be sponsored for self-pay settlement"
- [ ] `POST /relay` with `X-Settlement: self-pay` and an invalid tx hex: rejected with `400 INVALID_TRANSACTION`
- [ ] `POST /relay` with `X-Settlement: self-pay` where payment params don't match: rejected with `400 SETTLEMENT_VERIFICATION_FAILED`
- [ ] Rate limiting applies to self-pay requests the same as sponsored
- [ ] Dedup returns cached result on retry with same tx hex

All 23 existing unit tests pass (`npm test`).

### DIFF
--- a/src/endpoints/relay.ts
+++ b/src/endpoints/relay.ts
@@ -708,8 +708,9 @@ export class Relay extends BaseEndpoint {
     const storedReceipt = await receiptService.storeReceipt({
       receiptId,
       senderAddress: validation.senderAddress,
-      // No sponsor tx for self-pay — store the original tx for auditability
-      sponsoredTx: body.transaction,
+      // No sponsor tx for self-pay — sponsoredTx is intentionally undefined.
+      // The original transaction hex is available in the request log if needed.
+      sponsoredTx: undefined,
       fee: "0",
       txid: broadcastResult.txid,
       settlement,
@@ -738,7 +739,11 @@ export class Relay extends BaseEndpoint {
     return this.okWithTx(c, {
       txid: broadcastResult.txid,
       settlement,
-      sponsoredTx: null,
+      // sponsoredTx omitted (undefined not null): okWithTx's conditional spread skips
+      // falsy values so this field is absent from the JSON response — self-pay callers
+      // already have their tx and don't need it echoed back. Using undefined (not null)
+      // matches the optional field type and keeps JSON output consistent with omission.
+      sponsoredTx: undefined,
       receiptId: storedReceipt ? receiptId : undefined,
     });
   }
--- a/src/services/receipt.ts
+++ b/src/services/receipt.ts
@@ -15,7 +15,7 @@ export class ReceiptService {
   async storeReceipt(data: {
     receiptId: string;
     senderAddress: string;
-    sponsoredTx: string;
+    sponsoredTx?: string;
     fee: string;
     txid: string;
     settlement: SettlementResult;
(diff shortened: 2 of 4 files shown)


whoabuddy (approved):
## Review

**Approach:** Clean and non-invasive. The sponsored path is completely untouched — self-pay branches off early via the `X-Settlement: self-pay` header check and mirrors the sponsored flow step-by-step (rate limit, dedup, verify, broadcast, receipt, stats).

**What's good:**
- Proper rejection of sponsored txs submitted to the self-pay path
- `sponsoredTx` made optional in `PaymentReceipt` — correct since there's no sponsor sig
- Dedup and receipt storage work correctly with the optional field
- Address derivation uses the same hashMode-aware logic as the sponsored path

**Minor notes (not blocking):**
1. `handleSelfPay()` (~200 lines) largely duplicates the sponsored path logic — could be refactored to share a common `settleAndRecord()` helper in a follow-up
2. Fee recorded as `"0"` is accurate from the relay's perspective (relay didn't sponsor), worth a comment in docs
3. No test script for the new path — a `scripts/test-self-pay.ts` would be useful for manual testing

LGTM — merging as-is.

PR #163, opened 2026-03-12
feat: route RBF broadcasts through multi-node failover (closes #160)

### LINKED ISSUE #160: feat: route RBF broadcasts through multi-node failover path
## Context

During review of PR #158 (`fix/nonce-conflict-retry-signal`), arc0btc noted that `broadcastRbfForNonce` uses `@stacks/transactions` `broadcastTransaction` which goes directly to Hiro, bypassing the new multi-node failover in `broadcastAndConfirm`.

## Problem

If Hiro is partially degraded (reads work but broadcasts don't), RBF attempts will consume `MAX_RBF_ATTEMPTS` without any broadcast actually succeeding. The multi-node failover path added in the PR is not used for RBF broadcasts, so there is no automatic fallback to alternative nodes.

## Proposal

Route RBF broadcasts through the same multi-node failover path used by `broadcastAndConfirm`. This ensures that if the primary node is degraded for broadcasts, RBF attempts will automatically try alternative nodes before consuming retry attempts.

## References

- PR #158: https://github.com/aibtcdev/x402-sponsor-relay/pull/158

secret-mars:
## Summary

- Extracts the broadcast-only loop from `broadcastAndConfirm` into a new public `broadcastWithFailover` method on `SettlementService`
- Updates `broadcastRbfForNonce` in `nonce-do.ts` to call `broadcastWithFailover` instead of the single-node `@stacks/transactions.broadcastTransaction`
- Also updates `fillGapNonce` in `nonce-do.ts` for the same reason — gap-fill broadcasts were using the same single-node path
- No behaviour change for existing `broadcastAndConfirm` callers — it now delegates to `broadcastWithFailover` internally

## Problem

`broadcastRbfForNonce` (and `fillGapNonce`) called `broadcastTransaction()` from `@stacks/transactions`, which only routes to Hiro. If Hiro is partially degraded (reads work, broadcasts fail), all `MAX_RBF_ATTEMPTS` are consumed without ever trying fallback nodes. Stuck nonces remain unresolved for the full retry budget.

## Changes

- `src/services/settlement.ts`: Extract `broadcastWithFailover(transaction, sponsorNonceForLog?)` from the inner broadcast loop of `broadcastAndConfirm`. Returns `{ txid }` on success or `{ error, details, retryable, nonceConflict? }` on failure. Refactor `broadcastAndConfirm` to call it (DRY).
- `src/durable-objects/nonce-do.ts`: Remove `broadcastTransaction` import; add `SettlementService` import. Replace direct `broadcastTransaction` calls in `broadcastRbfForNonce` and `fillGapNonce` with `SettlementService.broadcastWithFailover` via a thin Logger adapter over `this.log()`.

## Test plan

- [x] TypeScript type check passes (`npm run check`)
- [x] All 23 existing tests pass (`npx vitest run`)
- [ ] Verify existing settlement flow still works end-to-end (`npm run test:relay`)
- [ ] Verify RBF broadcasts use fallback node when primary returns 5xx
- [ ] Verify `MAX_RBF_ATTEMPTS` budget is still respected (unchanged constant)

### DIFF
--- a/src/durable-objects/nonce-do.ts
+++ b/src/durable-objects/nonce-do.ts
@@ -2,7 +2,7 @@ import {
   makeSTXTokenTransfer,
 } from "@stacks/transactions";
 import { generateNewAccount, generateWallet } from "@stacks/wallet-sdk";
-import type { Env, LogsRPC } from "../types";
+import type { Env, Logger, LogsRPC } from "../types";
 import { getHiroBaseUrl, getHiroHeaders } from "../utils";
 import { SettlementService } from "../services/settlement";
 
@@ -412,6 +412,20 @@ export class NonceDO {
     }
   }
 
+  /**
+   * Returns a Logger adapter that routes SettlementService log calls through
+   * NonceDO's structured `log` method. Extracted to avoid copy-paste between
+   * fillGapNonce and broadcastRbfForNonce.
+   */
+  private makeSettlementLogger(): Logger {
+    return {
+      info: (msg: string, ctx?: Record<string, unknown>) => this.log("info", msg, ctx),
+      warn: (msg: string, ctx?: Record<string, unknown>) => this.log("warn", msg, ctx),
+      error: (msg: string, ctx?: Record<string, unknown>) => this.log("error", msg, ctx),
+      debug: (msg: string, ctx?: Record<string, unknown>) => this.log("debug", msg, ctx),
+    };
+  }
+
   private getStateValue(key: string): number | null {
     const rows = this.sql
       .exec<{ value: number }>(
@@ -686,14 +700,10 @@ export class NonceDO {
 
       // Use SettlementService.broadcastWithFailover so gap-fill broadcasts also
       // benefit from multi-node failover when Hiro is degraded.
-      const doLogger = {
-        info: (msg: string, ctx?: Record<string, unknown>) => this.log("info", msg, ctx),
-        warn: (msg: string, ctx?: Record<string, unknown>) => this.log("warn", msg, ctx),
-        error: (msg: string, ctx?: Record<string, unknown>) => this.log("error", msg, ctx),
-        debug: (msg: string, ctx?: Record<string, unknown>) => this.log("debug", msg, ctx),
-      };
-      const settlement = new SettlementService(this.env, doLogger);
-      const result = await settlement.broadcastWithFailover(tx, null);
+      // SettlementService is lightweight (stores refs to env and logger only),
+      // so constructing it per call has no meaningful overhead.
+      const settlement = new SettlementService(this.env, this.makeSettlementLogger());
+      const result = await settlement.broadcastWithFailover(tx, gapNonce);
 
       if ("txid" in result) {
         return result.txid;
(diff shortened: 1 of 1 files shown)


arc0btc (commented):
Routes RBF and gap-fill broadcasts through the same multi-node failover path as normal settlements — addresses a real gap we hit operationally when Hiro is partially degraded.

**What works well:**
- The extraction of `broadcastWithFailover` is clean and well-motivated. Previously, `broadcastRbfForNonce` and `fillGapNonce` called `@stacks/transactions.broadcastTransaction` directly, bypassing the failover logic entirely. Under a degraded Hiro, all `MAX_RBF_ATTEMPTS` were consumed against a single failing endpoint.
- The discriminated union return type (`{ txid }` | `{ error, details, retryable, nonceConflict? }`) is explicit and readable. Type-safe nonce conflict handling (`result.nonceConflict`) is a clear improvement over the previous `as unknown as { reason?: string }` cast.
- `broadcastAndConfirm` delegating to `broadcastWithFailover` is the right DRY call — one broadcast path to maintain.

**[question] Retry multiplication in the RBF loop**

`broadcastWithFailover` contains its own inner retry loop: up to `BROADCAST_MAX_ATTEMPTS` per node × N nodes from `getBroadcastTargets`. `broadcastRbfForNonce` calls this from inside a `MAX_RBF_ATTEMPTS` outer loop. Total network attempts per stuck nonce = `MAX_RBF_ATTEMPTS × BROADCAST_MAX_ATTEMPTS × nodeCount`.

If `MAX_RBF_ATTEMPTS=10`, `BROADCAST_MAX_ATTEMPTS=3`, and there are 2 nodes, that's 60 broadcast calls per nonce — a significant increase from the previous 10. What are the actual values, and is this multiplication intentional? 4xx short-circuits immediately so this matters most when nodes are consistently 5xx-ing — exactly the degraded Hiro scenario this is trying to solve. Just want to confirm the math was considered before the test plan item about `MAX_RBF_ATTEMPTS` budget is checked off.

**[question] `sponsorNonceForLog: null` for gap-fill**

In `fillGapNonce`, the gap nonce is known (`gapNonce`) but `null` is passed to `broadcastWithFailover`. The param is log-only so this doesn't affect behavior, but passing `gapNonce` would make broadcast log lines in `SettlementService` attributable to the specific nonce being filled. Worth considering.

**[nit] Logger adapter duplication**

The `doLogger` adapter block (4 lines mapping `info/warn/error/debug` to `this.log`) is copy-pasted identically in `fillGapNonce` and `broadcastRbfForNonce`. Could be a private helper `#makeSettlementLogger(): Logger` on `NonceDO` to keep it in one place.

**[nit] Instantiation per call**

`new SettlementService(this.env, doLogger)` is constructed fresh on every RBF attempt. If `SettlementService` is lightweight (just stores refs), this is fine. If it parses config or sets up clients in the constructor, it could be worth moving to a lazy field on the DO. Probably fine — just flagging.

**Operational context:** We're running v1.18.0 of this relay and we've hit the NONCE_CONFLICT cascade directly — nonce pool exhaustion when Hiro was degraded was exactly this scenario. This change would have helped. The failover path already exists and works for normal settlements; extending it to RBF is the right call.

TypeScript check and unit tests passing is a solid baseline. The unchecked end-to-end items in the test plan are the main gap before merge.

PR #194, opened 2026-03-23
fix: headroom-aware wallet selection + soft-reject at low headroom

### LINKED ISSUE #193: fix: use local nonce tracking to load-balance across wallets and avoid TooMuchChaining
## Context

On 2026-03-22, 5 transactions were rejected with `TooMuchChaining` during burst traffic between 10-11am. The Stacks mempool enforces a limit of 25 chained unconfirmed transactions per address. Currently, the relay only discovers this limit when Hiro API rejects the broadcast — too late.

## Problem

When a single sponsor wallet accumulates close to 25 pending txs and a burst of new requests arrives, the relay keeps submitting to the same wallet until the node rejects with `TooMuchChaining`. This wastes time and causes downstream payment failures.

## Proposal

Since we already maintain local nonce state per wallet (post #165 refactor), we have the data to solve this locally:

1. **Track pending chain depth per wallet** — count of broadcast-but-unconfirmed txs (nonces between last confirmed and current tip)
2. **Load-balance new requests** — when selecting a sponsor wallet, prefer the one with the most chain headroom (`25 - pending_depth`), not just round-robin
3. **Soft-reject at threshold** — if all wallets have `<= 2` headroom, return a 503 with `Retry-After` rather than broadcasting into a guaranteed rejection
4. **Expand wallet pool if needed** — if sustained traffic keeps all wallets near capacity, adding more sponsor addresses is straightforward

This is strictly better than relying on Hiro API rejections because:
- No wasted broadcast round-trips
- Predictable backpressure before failures occur  
- Agents see degradation in `/health` (see #191) and can queue locally

## Expected traffic

Expecting a significant registration and messaging spike on 2026-03-23. Current wallet count should handle normal load, but burst patterns like Mar 22 would benefit from chain-depth-aware routing.

## References

- #191 — surface nonce pool health in /health (will include chain depth metrics)
- #187 — nonce pool capacity expansion
- Logs: `TooMuchChaining` rejections at logs.aibtc.com on 2026-03-22 10:00-11:20 UTC

gregoryford963-sys:
Fixes #193.

## Problem

The relay used strict round-robin, picking the **first** wallet under `CHAINING_LIMIT`. During burst traffic (like the Mar 22 incident), multiple concurrent requests could land on the same nearly-full wallet before any confirmed, triggering `TooMuchChaining` rejections.

### 1. Headroom-aware wallet selection

The assignment loop now collects **all** eligible wallets in one pass instead of stopping at the first one under the limit. After the scan, it selects the wallet with the **most available headroom** (`CHAINING_LIMIT - reservedCount`):

```typescript
// Before: first-eligible round-robin
if (this.ledgerReservedCount(walletIndex) < CHAINING_LIMIT) {
  selectedWalletIndex = walletIndex;
  break;
}

// After: collect all eligible, pick most headroom
eligibleWallets.push({ walletIndex, headroom: CHAINING_LIMIT - reserved });
// ... after loop:
eligibleWallets.sort((a, b) => b.headroom - a.headroom);
selectedWalletIndex = eligibleWallets[0].walletIndex;
```

This spreads load toward the wallet with the most room to absorb additional requests.

### 2. Soft-reject at low headroom

New constant: `SOFT_REJECT_HEADROOM_THRESHOLD = 2`

If even the best wallet has headroom ≤ 2 slots, throws `LowHeadroomError` → **503** with `Retry-After` header instead of assigning into a pool that will likely hit `TooMuchChaining` in the next burst:

```json
HTTP 503 Low Headroom
Retry-After: 10

{
  "code": "LOW_HEADROOM",
  "maxHeadroom": 1,
  "retryAfterSeconds": 10
}
```

Agents receiving 503 can queue locally and retry — no wasted broadcast round-trips.

## Scope

- Single file: `src/durable-objects/nonce-do.ts`
- Existing degraded-wallet fallback and `ChainingLimitError` (429) unchanged
- No changes to existing interfaces or external API (adds new 503 response code)

## Test plan

- [ ] Single wallet, headroom > 2: selects it (unchanged behavior)
- [ ] Multi-wallet: selects wallet with most headroom, not first-eligible
- [ ] All wallets at headroom ≤ 2: returns 503 + `Retry-After`
- [ ] All wallets at CHAINING_LIMIT: still returns 429 (ChainingLimitError path unchanged)


gregoryford963-sys:
Addressed the review feedback in the follow-up commit (7b6a0b2):

- **Scale threshold with limit**: `SOFT_REJECT_HEADROOM_THRESHOLD` is now `Math.ceil(CHAINING_LIMIT * 0.1)` — stays at 2 for limit=20 but adapts automatically if the limit changes
- **Fix retryAfterSeconds formula**: backoff is now proportional to pool pressure (`(CHAINING_LIMIT - maxHeadroom) / 2 + 5`) rather than the near-constant threshold delta that always yielded ~6s
- **Remove dead computation**: dropped the `totalMempoolDepth` accumulation in the success path — that value was never read after wallet selection
- **Remove maxHeadroom from 503 body**: callers only need `retryAfterSeconds` to back off correctly; no need to expose internal pool capacity

Ready for re-review / merge when convenient.

PR #247, opened 2026-03-27
feat: nonce pool hardening — malformed payload rejection, dispatch queue, agent queue management

### LINKED ISSUE #241: feat: improve nonce pool resilience under burst traffic
## Context

During a nonce storm on 2026-03-26 (12:49–1:10 PM PST), 9 agents simultaneously hit `/api/inbox` on the landing page, creating a cascade of `ConflictingNonceInMempool` errors. The relay's detection mechanisms (circuit breaker, cascade detection, quarantine) all fired correctly, but the **recovery path is too passive** — it waits for agents to retry rather than self-healing inline.

**Evidence from log triage:**
- 480 errors on `aibtc-landing`, almost all NONCE_CONFLICT + RELAY_ERROR
- 781 warnings on `x402-relay` (quarantines, cascades, stale reconciliations)
- Wallets 5/6/8/9 accumulated mempool backlogs that took hours to drain
- Agents retried every ~7s despite `retryAfter: 30` — amplifying the cascade
- Zero relay errors all day — the relay detected everything, just didn't react fast enough

Related: https://github.com/aibtcdev/landing-page/issues/522

### 1. Inline resync + retry on `/relay` endpoint (HIGH priority)

**Problem:** `/relay` only schedules a *delayed* resync on nonce conflict (`relay.ts:424`), then returns 409 to the agent. The agent retries seconds later, hits another stale nonce, repeat. Meanwhile `/settle` does an *inline* resync + immediate retry (`settle.ts:495-540`) and often succeeds on the second attempt.

**Fix:** When `/relay` gets `nonceConflict`, do `await resyncNonceDO()` + re-sponsor + re-broadcast (same pattern as `/settle`). Fall back to 409 only if the retry also fails.

**Impact:** Would have prevented the majority of the 326 NONCE_CONFLICT errors agents saw during the storm.

### 2. Per-sender dedup window (HIGH priority)

**Problem:** Agents retry the same inbox destination every ~7s during conflicts. Each retry goes through the full pipeline: nonce assignment → broadcast → conflict → quarantine. This amplifies cascades.

**Fix:** Add a lightweight check early in the `/relay` pipeline:
- If sender X got a conflict <`retryAfter` seconds ago, short-circuit with cached 409 + remaining cooldown
- Don't assign a new nonce, don't hit the mempool
- Key on `senderSignerHash` (already extracted for sender-nonce checks)
- Could use KV with TTL matching retryAfter

**Impact:** Prevents the "assign nonce → broadcast → conflict → quarantine" cycle from repeating per-agent per-7-seconds.

### 3. Dynamic `retryAfter` based on pool pressure (MEDIUM priority)

**Problem:** Currently hardcoded `retryAfter: 30` on every NONCE_CONFLICT (`relay.ts:431`). But pool state varies — sometimes 8/10 wallets are clean and retry could succeed in 10s; other times all wallets are contended and 30s is too short.

**Fix:** Calculate retryAfter from observable pool state:
- Query pool pressure (already computed in `nonce_pool_pressure` logs)
- Low contention (1-2 wallets affected): `retryAfter: 10`
- Medium (3-5 wallets): `retryAfter: 30` (current default)
- High (6+ wallets or cascade active): `retryAfter: 60`

The `LowHeadroomError` already calculates dynamic retry (`nonce-do.ts:414`) — extend that pattern.

### 4. Cascade-aware RBF timing (MEDIUM priority)

**Problem:** `STUCK_TX_AGE_MS = 15 * 60 * 1000` (15 min, `nonce-do.ts:314`). During a burst, txs get stuck but the relay waits 15 minutes before RBF kicks in. By then wallets have accumulated large backlogs.

**Fix:** Make RBF age threshold responsive to pool state:
- Normal: 15 min (current)
- When `cascade_detected` fires: drop to 5 min
- When circuit breaker is open: drop to 3 min

The cascade/circuit-breaker signals are already emitted — just need to feed them into the RBF eligibility check in reconciliation (`nonce-do.ts:2897`).

### 5. Bump gap-fill limit during cascades (LOW priority)

**Problem:** `MAX_GAP_FILLS_PER_ALARM = 5` (`nonce-do.ts:306`). With 60s alarm intervals, a wallet with 10 gaps takes 2 cycles (2 min) to clear. During a cascade, this is too slow.

**Fix:** Scale to 10-15 gap-fills per alarm when cascade is active. Gap-fills are cheap (1 uSTX + 30k fee) — the cost of a larger batch is negligible vs the cost of prolonged pool contention.

### 6. Aggressive reconciliation during cascades (LOW priority)

**Problem:** Alarm interval scales between 60s (active) and 5min (idle) but doesn't respond to cascade events. When `cascade_detected` fires, the pool is in its worst state but reconciliation runs at the same cadence.

**Fix:** When cascade is active, drop alarm interval to 15-30s. Return to normal cadence when cascade clears (quarantine count drops below threshold).

## Implementation Order

| # | Change | Files | Effort |
|---|--------|-------|--------|
| 1 | Inline resync + retry on `/relay` | `src/endpoints/relay.ts` | Medium |
| 2 | Per-sender dedup window | `src/endpoints/relay.ts`, possibly `src/services/sender-nonce.ts` | Low |
| 3 | Dynamic retryAfter | `src/endpoints/relay.ts`, `src/durable-objects/nonce-do.ts` | Low |
| 4 | Cascade-aware RBF timing | `src/durable-objects/nonce-do.ts` | Low |
| 5 | Bump gap-fill limit | `src/durable-objects/nonce-do.ts` | Trivial |
| 6 | Aggressive reconciliation alarm | `src/durable-objects/nonce-do.ts` | Low |

Items 1 + 2 are the highest-leverage changes and could be shipped together. Items 3-6 are tuning knobs that can follow.

## Test Plan

- [ ] Simulate burst: 10 concurrent `/relay` requests from different senders to same sponsor wallet
- [ ] Verify inline resync + retry succeeds on first conflict (item 1)
- [ ] Verify dedup window short-circuits repeat requests within cooldown (item 2)
- [ ] Verify retryAfter scales with pool pressure (item 3)
- [ ] Verify RBF triggers faster when cascade is active (item 4)
- [ ] Monitor wallet drain time after deploying — target: <5 min to clear backlogs vs current ~hours

### LINKED ISSUE #245: bug: wallets 6-9 not draining — pending txs stuck 20+ hours
## Context

During log triage on 2026-03-26 and 2026-03-27, wallets 6-9 in the nonce pool have accumulated pending transactions that are not confirming or being cleared by RBF. Wallet 6 has been at the same pending state for **20+ hours**.

Related: #241 (nonce pool resilience improvements — items 4-6 specifically)

### Wallet pool comparison (2026-03-26 15:35 PST → 2026-03-27 ~05:30 PST)

| Wallet | Mar 26 15:35 | Mar 27 05:30 | Change |
|--------|-------------|-------------|--------|
| 0-5 | Mixed pending | Clean | Drained ✓ |
| 6 | 9 pending, 1 RBF, 18 mempool | 9 pending, 1 RBF, 18 mempool | **No change in 20hrs** |
| 7 | 0 pending, 4 mempool | 21 pending, 1 RBF, 26 mempool | **Growing** |
| 8 | 1 diverge, 9 pending | 25 pending, 1 RBF, 25 mempool | **Growing** |
| 9 | 18 pending, 22 mempool | 18 pending, 1 RBF, 22 mempool | **Same — not draining** |

### Key observations

1. **RBF is firing** (1 candidate per wallet per reconciliation) but only processes 1 tx at a time — insufficient to drain backlog
2. **All pending are `pending_agree`** (no divergence) — ledger and Hiro agree these are in mempool, they're just not confirming
3. **Wallets 0-5 drained fine** — whatever is different about 6-9 is specific to those wallets
4. **No circuit breaker or cascade events** — the relay doesn't see this as a problem state

## Expected Behavior

Pending transactions that have been in mempool for 20+ hours should be aggressively RBF'd or gap-filled. The pool should be self-healing within minutes, not days.

## Root Cause Hypotheses

1. **RBF throughput too low:** With `MAX_GAP_FILLS_PER_ALARM = 5` and 60s alarm interval, a wallet with 25 pending txs would take 5 cycles (5 min) to clear — but only if they're all RBF candidates. The 15-minute `STUCK_TX_AGE_MS` threshold means newly-stuck txs wait before becoming eligible.

2. **`pending_agree` prevents RBF:** The reconciliation code only marks a tx as an RBF candidate when it's **missing from Hiro's mempool** AND >15min old. If Hiro still shows the tx in mempool (`pending_agree`), it never becomes an RBF candidate — even if it's been pending for 20 hours.

3. **Mempool congestion:** These txs may be genuinely stuck in the Stacks mempool with low fees. The relay's gap-fill fee (30k uSTX) and RBF fee (90k uSTX) may be below the current fee market.

## Suggested Investigation

1. Check if the `pending_agree` txs on wallets 6-9 are actually in the Stacks mempool (query Hiro directly for those nonces)
2. If they are in mempool: the relay needs a "pending too long" → force RBF path (not just "missing from mempool")
3. If they aren't in mempool: Hiro's mempool index may be stale — relay should cross-check with a second node
4. Check fee levels on the stuck txs vs current fee market

### Immediate: Add `pending_too_long` RBF path

In `reconcileNonceForWallet()`, add a new verdict alongside the existing `pending_agree`:

```
if (ledger says broadcasted) AND (Hiro says in mempool) AND (age > 60 minutes):
  → verdict: "pending_too_long" → add to RBF candidates
```

This catches the case where a tx is genuinely stuck in mempool but Hiro still sees it.

### From #241: Cascade-aware tuning

These changes from #241 would also help:
- **Item 4:** Reduce `STUCK_TX_AGE_MS` to 5min during cascades
- **Item 5:** Increase `MAX_GAP_FILLS_PER_ALARM` to 10-15 during cascades
- **Item 6:** Drop alarm interval to 15-30s during cascades

### Monitoring: Alert on stale pending

Add a health check that flags when any wallet has had `pending_agree > 0` for more than 30 minutes. This should be visible on the `/nonce/stats` endpoint.

## Test Plan

- [ ] Verify `pending_too_long` threshold triggers RBF on txs >60min in mempool
- [ ] Verify wallets drain within 10 minutes under simulated backlog
- [ ] Verify RBF fee is competitive with current fee market
- [ ] Monitor wallet pool after deploy — all wallets should be at 0 pending within 1 hour
- [ ] Load test: simulate 20 concurrent txs per wallet, verify full drain < 15 min

### LINKED ISSUE #246: feat: reject malformed transaction payloads at validation layer
## Context

On 2026-03-27, 4 requests hit the relay with payloads that failed deserialization:

```
2x "Could not parse 48 as TransactionVersion"  — hex byte 0x30 (ASCII '0'), not a valid Stacks version
2x "Could not parse 0 as AuthType"              — empty/zeroed auth field
```

Currently these fail at `deserializeTransaction()` in the sponsor pipeline, generate a warn log, and return an error. While the relay handles them correctly today, at 100x volume this becomes a concern:

1. **Wasted compute:** Each malformed payload goes through auth middleware, rate limit checks, and JSON parsing before failing at deserialization
2. **Noisy logs:** Warns that aren't actionable mix with real operational warnings
3. **No sender accountability:** The relay doesn't track which senders are sending bad payloads

### 1. Early validation before sponsor pipeline

Add a lightweight tx hex validation step immediately after request parsing, before nonce assignment or sponsor service calls:

```typescript
// Quick check: valid Stacks tx starts with version byte 0x00 (mainnet) or 0x80 (testnet)
// and auth type byte 0x04 (standard) or 0x05 (sponsored)
function quickValidateStxTxHex(hex: string): boolean {
  const bytes = hexToBytes(hex);
  if (bytes.length < 2) return false;
  if (bytes[0] !== 0x00 && bytes[0] !== 0x80) return false; // TransactionVersion
  if (bytes[1] !== 0x04 && bytes[1] !== 0x05) return false; // AuthType
  return true;
}
```

Return 400 `INVALID_TRANSACTION` immediately — don't enter the sponsor pipeline.

### 2. Track bad-payload senders

Log the sender address (if extractable) or IP with the rejection. If the same source sends 3+ malformed payloads in 10 minutes, temporarily block at the rate limit layer.

### 3. Distinguish from client errors in logging

Change from `WARN "Failed to deserialize transaction"` to:
- `INFO "Rejected malformed transaction"` (expected, not worth alerting on)
- Include: raw hex length, first 4 bytes, source identifier

## Evidence

```json
// TransactionVersion 48 = ASCII '0' — someone sent a string starting with "0x..." 
// as literal ASCII instead of hex-decoded bytes
{"error": "Could not parse 48 as TransactionVersion"}

// AuthType 0 = no auth — possibly a raw BTC transaction or random bytes
{"error": "Could not parse 0 as AuthType"}
```

## Test Plan

- [ ] Send valid Stacks tx → accepted as before
- [ ] Send hex starting with `0x30` (ASCII '0') → rejected 400 before sponsor pipeline
- [ ] Send empty/zeroed payload → rejected 400
- [ ] Send raw BTC tx hex → rejected 400
- [ ] Verify rejected payloads don't trigger nonce assignment or broadcast
- [ ] Verify 3+ rejections from same source triggers temporary block

whoabuddy:
## Summary

- **Reject malformed payloads early (#246):** O(1) hex pre-validation (`preValidateTxHex`) checks version byte, auth type, and hex validity before expensive `deserializeTransaction()`. IP-based rate limiting blocks repeat offenders (3+ malformed payloads in 10 min → 429). Expected-rejection logs downgraded from WARN to INFO.
- **Dispatch queue with flush-and-replay (#245):** New `dispatch_queue` and `replay_buffer` SQL tables in NonceDO. `runFlushAndReplayCycle()` detects stuck dispatched entries, flushes sponsor nonce slots with self-transfers, and moves sender txs to replay buffer for re-sponsoring. Queue stats wired into `/nonce/state`, `/stats`, and pool health.
- **Close #241 items 4-6:** Verified all 6 items shipped in PR #242 (e9ba970), posted confirmation comment on the issue.
- **Agent queue visibility:** New `GET /queue/:senderAddress` and `DELETE /queue/:senderAddress/:walletIndex/:sponsorNonce` endpoints with SIP-018 auth (`queue-read`/`queue-cancel` actions). Agents can see their queue position and cancel their own txs. AX discovery docs updated with queue management topic.

## Test plan

- [ ] `npm run check` — TypeScript clean
- [ ] `npm run deploy:dry-run` — build succeeds
- [ ] `npm run test:relay` — existing relay flow unchanged
- [ ] `npm run test:sponsor` — existing sponsor flow unchanged
- [ ] `npm run test:settle` — existing settle flow unchanged
- [ ] `npm run test:queue` — new queue read/cancel endpoints
- [ ] Submit malformed hex to `/relay` — verify `MALFORMED_PAYLOAD` error before deserialization
- [ ] Submit 3+ malformed payloads from same IP — verify 429 response
- [ ] Verify `/nonce/state` includes `queueDepth` and `replayBufferDepth` per wallet

Closes #245, closes #246

### DIFF
--- a/src/durable-objects/nonce-do.ts
+++ b/src/durable-objects/nonce-do.ts
@@ -663,13 +663,14 @@ export class NonceDO {
       `INSERT OR REPLACE INTO dispatch_queue
          (wallet_index, position, sender_tx_hex, sender_address, sender_nonce,
           sponsor_nonce, state, queued_at, dispatched_at, confirmed_at)
-       VALUES (?, ?, ?, ?, ?, ?, 'queued', ?, NULL, NULL)`,
+       VALUES (?, ?, ?, ?, ?, ?, 'dispatched', ?, ?, NULL)`,
       walletIndex,
       position,
       senderTxHex,
       senderAddress,
       senderNonce,
       sponsorNonce,
+      now,
       now
     );
   }
@@ -3161,7 +3162,8 @@ export class NonceDO {
         .toArray();
 
       if (stuckEntries.length === 0) {
-        return { flushed: 0, replayBufferDepth: 0 };
+        const currentDepth = this.getReplayBufferDepth(walletIndex);
+        return { flushed: 0, replayBufferDepth: currentDepth };
       }
 
       // Batch flush detection: log a high-visibility event when many slots are stuck
@@ -3367,8 +3369,10 @@ export class NonceDO {
             continue;
           }
 
-          // Reserve the nonce in the ledger
+          // Reserve the nonce in the ledger and advance head to mirror normal assignment path
           this.ledgerAssign(targetWallet.walletIndex, freshNonce);
+          this.ledgerAdvanceWalletHead(targetWallet.walletIndex, freshNonce + 1);
+          this.updateAssignedStats(freshNonce);
 
           // Deserialize and re-sponsor the sender tx
           const cleanHex = entry.sender_tx_hex.replace(/^0x/i, "");
--- a/src/endpoints/queue-cancel.ts
+++ b/src/endpoints/queue-cancel.ts
@@ -15,7 +15,7 @@ import {
  *
  * State transitions:
  *   queued → deleted immediately
- *   dispatched → set to 'replaying' (signals the alarm to flush the sponsor nonce slot)
+ *   dispatched → sponsor nonce flushed immediately (best-effort via RBF), then deleted
  *   replaying → deleted immediately
  *   replay_buffer → deleted (matched by original_sponsor_nonce)
  *
@@ -31,7 +31,7 @@ export class QueueCancel extends BaseEndpoint {
       "The recovered signer address must match the :senderAddress URL parameter.\n\n" +
       "State transitions:\n" +
       "- queued → deleted immediately\n" +
-      "- dispatched → transitioned to 'replaying' (sponsor nonce slot will be flushed in next alarm cycle)\n" +
+      "- dispatched → sponsor nonce flushed immediately (best-effort via RBF), then deleted\n" +
       "- replaying → deleted immediately\n" +
       "- replay_buffer entry (matched by original_sponsor_nonce) → deleted",
     request: {
@@ -101,8 +101,10 @@ export class QueueCancel extends BaseEndpoint {
               type: "object" as const,
               properties: {
                 success: { type: "boolean" as const, example: false },
+                requestId: { type: "string" as const, format: "uuid" },
                 error: { type: "string" as const },
                 code: { type: "string" as const, example: "QUEUE_ACCESS_DENIED" },
+                retryable: { type: "boolean" as const, example: false },
               },
             },
           },
--- a/src/middleware/malformed-tracker.ts
+++ b/src/middleware/malformed-tracker.ts
@@ -4,8 +4,11 @@
  * Senders who submit 3+ malformed payloads within a 10-minute window
  * are temporarily blocked for the remainder of the window.
  *
- * A single shared map is used across all endpoints so that an attacker
- * cannot bypass the threshold by rotating between /relay, /sponsor, and /settle.
+ * can be temporarily blocked for the remainder of the window by callers that
+ * enforce the result of this tracker.
+ *
+ * A single shared map is used so that any endpoints that consult this tracker
+ * share the same malformed-attempt counts for a given IP.
  */
 
 /** Duration of the tracking window in milliseconds (10 minutes). */
--- a/src/routes/discovery.ts
+++ b/src/routes/discovery.ts
@@ -1733,17 +1733,21 @@ Returns the agent's current pending transaction queue across all sponsor wallets
 
 action: "queue-read"
 
-Request body:
+This endpoint uses SIP-018 auth via the \`X-SIP018-Auth\` HTTP header (not a JSON body).
+The header value MUST be a JSON string with the following shape:
+
 {
-  "auth": {
-    "signature": "0x...",           // RSV hex SIP-018 signature
-    "message": {
-      "action": "queue-read",       // must be exactly "queue-read"
-      "nonce": "1708099200000",     // unix ms timestamp (replay protection)
-      "expiry": "1708185600000"     // must be in the future
-    }
+  "signature": "0x...",           // RSV hex SIP-018 signature
+  "message": {
+    "action": "queue-read",       // must be exactly "queue-read"
+    "nonce": "1708099200000",     // unix ms timestamp (replay protection)
+    "expiry": "1708185600000"     // must be in the future
   }
 }
+
+Example:
+
+X-SIP018-Auth: {"signature":"0x...","message":{"action":"queue-read","nonce":"1708099200000","expiry":"1708185600000"}}
 
 ### Response (200)
 
@@ -1772,9 +1776,8 @@ Cancels a queued, dispatched, or replaying sponsored transaction.
 ### State Transitions
 
 - queued → deleted immediately from the dispatch queue
-- dispatched → transitioned to 'replaying' (the sponsor nonce slot is marked
-  for flushing in the next alarm cycle; the sender tx will be re-sponsored
-  with a fresh nonce and can be resubmitted)
+- dispatched → sponsor nonce flushed immediately (best-effort) and the
+  corresponding dispatch_queue row deleted (no 'replaying' transition)
 - replaying → deleted immediately from the dispatch queue
 - replay_buffer entry (matched by original_sponsor_nonce) → deleted
 


arc0btc (approved):
## Re-review — 3e9e968 (all items resolved)

All five remaining items from my previous re-review are addressed correctly:

### Blocking fix — verified

**replay_buffer orphan on `replaying` cancel** — Fixed. Both `dispatch_queue` and `replay_buffer` rows are now deleted atomically. The `original_sponsor_nonce` join key is correct. This was the last blocker.

### Non-blocking fixes — verified

| Item | Status |
|------|--------|
| `replayed` → `replayBufferDepth` | Renamed in return type, variable, log field (`flush_replay_buffer_depth`), and comments clarify it's total depth |
| 403 passes through DO error message | `doBody?.error ?? "Access denied..."` fallback is clean |
| `LIMIT 100` on queue-sender queries | Applied to both `dispatch_queue` and `replay_buffer` SELECT statements |
| `string-ascii 10` → `string-ascii 12` | Both discovery doc locations updated, comment lists all four action strings |

LGTM. Ready to merge.

PR #268, opened 2026-03-29
fix(nonce): backward probe activates after failed gap-fills (closes #262)

### LINKED ISSUE #262: fix(nonce): backward probe should activate for failed gap_fills
## Problem

The backward ghost probe (PR #261) only activates when `flush-wallet` has an **empty forward range** (`chainFrontier === assignmentHead`, no gaps). But the scenario it was designed for — ghost nonces the node won't release — creates a **non-empty** forward range because Hiro reports them as `detected_missing_nonces`.

### Current behavior

1. `flush-wallet walletIndex=4 probeDepth=25` runs
2. Relay syncs from Hiro, discovers gaps at 702-717 → forward range is non-empty
3. Forward `gap_fill` runs, fills some nonces, but ghosts at 702/707/709/710/715 are rejected ("broadcast rejected or already occupied")
4. Backward probe **never activates** because forward range was non-empty
5. Repeated calls produce identical results — ghosts never get RBF'd at 90k fee

### Expected behavior

When `flush-wallet` completes with `failed` nonces (rejected broadcasts) and `probeDepth` is set, the backward probe should enqueue the failed nonces into `probe_queue` for alarm-driven RBF at `RBF_FEE`. These are exactly the ghost entries the probe was designed to evict.

### Live example (wallet 4)

```
Chain state (Hiro): lastExecuted=701, missing=[702,704,708,709,713,714], mempool=9 txs
Relay state: chainFrontier=717, assignmentHead=717, gaps=[], healthy=true
flush-wallet result: 10 filled, 5 failed (702,707,709,710,715 — ghosts)
probeQueue: empty (never activated)
```

### Suggested fix

After the forward flush loop, if `probeDepth > 0` and `failed.length > 0`, enqueue each failed nonce into `probe_queue` for alarm-driven RBF processing — same path as the empty-range probe but seeded with known-stuck nonces instead of backward enumeration.

tfireubs-ui:
## Problem

`flush-wallet` with `probeDepth` set never enqueues failed nonces into `probe_queue` because the backward probe only activates when the **forward range is empty** (`flushStart >= flushEnd`). But the scenario the probe was designed for — ghost nonces that Hiro reports as `detected_missing_nonces` — creates a **non-empty** forward range. The forward `gap_fill` runs, fails on the ghosts ("broadcast rejected or already occupied"), and the probe never fires.

Closes #262.

## Root Cause

From `handleFlushWallet` (line ~6595):

```typescript
// Backward probe only fires on empty forward range
if (flushStart >= flushEnd && probeDepth && probeDepth > 0 && ...) {
  // enqueue probe range
  return jsonResponse({ mode: "backward_probe_enqueued", ... });
}
// Forward flush runs...
// failedNonces accumulated but never enqueued for probe
return jsonResponse({ filled, failed: failedNonces, ... });
```

## Fix

After the forward flush loop completes (Step 3 — advance wallet head), if `probeDepth > 0` and `failedNonces.length > 0`, insert each failed nonce into `probe_queue` using `INSERT OR IGNORE`:

```typescript
if (probeDepth && probeDepth > 0 && failedNonces.length > 0) {
  for (const { nonce } of failedNonces) {
    this.sql.exec(
      `INSERT OR IGNORE INTO probe_queue (wallet_index, nonce, state, created_at)
       VALUES (?, ?, 'pending', ?)`,
      walletIndex, nonce, now
    );
  }
}
```

Key design choices:
- **`INSERT OR IGNORE` (not DELETE + INSERT)**: the empty-range path wipes the queue before bulk-inserting a backward range. Here we only add specific known-stuck nonces, preserving any pending probe entries from a prior cycle.
- **No early return**: the flush response still includes `filled` and `failed`, with an additional `probeEnqueued` count and `probeNote` when nonces were queued.
- **`flush_wallet_complete` log** now includes `probeEnqueued` for observability.

## Test Plan

- [ ] `npm run check` passes
- [ ] `flush-wallet walletIndex=N probeDepth=25` with known ghost nonces → response includes `probeEnqueued > 0` and `probeNote`
- [ ] After alarm fires: `GET /nonce/state` shows probe_queue entries transitioning from `pending` → `replaced` (or `conflict`/`rejected`)
- [ ] Ghost nonces evicted from mempool after 1-2 alarm ticks

### DIFF
--- a/src/durable-objects/nonce-do.ts
+++ b/src/durable-objects/nonce-do.ts
@@ -2115,7 +2115,12 @@ export class NonceDO {
 
   /**
    * Broadcast a gap-fill STX transfer for a specific nonce.
-   * Returns the txid on success, null if the nonce is already occupied or on error.
+   * Returns { ok: true, txid } on success.
+   * Returns { ok: false, reason } on failure, where `reason` is the broadcast
+   * error code (e.g. "ConflictingNonceInMempool") or a synthetic code
+   * ("gap_fill_error" for transport failures, "gap_fill_rejected" for other 4xx).
+   * Callers that need to distinguish slot occupancy from transient failures should
+   * check whether reason === "ConflictingNonceInMempool".
    * Amount: 1 uSTX. Fee: 30,000 uSTX (RBF-capable). Memo: gap-fill-{nonce}.
    */
   private async fillGapNonce(
@@ -2123,7 +2128,7 @@ export class NonceDO {
     gapNonce: number,
     privateKey: string,
     feeOverride?: bigint
-  ): Promise<string | null> {
+  ): Promise<{ ok: true; txid: string } | { ok: false; reason: string }> {
     const fee = feeOverride ?? GAP_FILL_FEE;
     // Track gap-fill attempts for this nonce (used for fee escalation).
     // Uses INSERT ON CONFLICT so the counter is created even when no nonce_intents
@@ -2151,7 +2156,7 @@ export class NonceDO {
       });
       const result = await this.broadcastRawTx(tx, "gap_fill");
       if (result.ok) {
-        return result.txid;
+        return { ok: true, txid: result.txid };
       }
       if (result.reason === "ConflictingNonceInMempool") {
         // Nonce already occupied — update ledger to prevent re-queuing on next alarm cycle.
@@ -2212,7 +2217,7 @@ export class NonceDO {
             );
           }
         } catch { /* fail-open */ }
-        return null;
+        return { ok: false, reason: "ConflictingNonceInMempool" };
       }
       this.log("warn", "gap_fill_rejected", {
         walletIndex,
@@ -2221,14 +2226,14 @@ export class NonceDO {
         reason: result.reason,
         body: result.body,
       });
-      return null;
+      return { ok: false, reason: result.reason ?? "gap_fill_rejected" };
     } catch (e) {
       this.log("warn", "gap_fill_error", {
         walletIndex,
         nonce: gapNonce,
         error: e instanceof Error ? e.message : String(e),
       });
-      return null;
+      return { ok: false, reason: "gap_fill_error" };
     }
   }
 
@@ -5513,18 +5518,18 @@ export class NonceDO {
           // Fee escalation: conflict retries use GAP_FILL_FEE + prior attempts (+1 uSTX each)
           const escalatedFee = this.computeEscalatedFee(walletIndex, gapNonce);
           const feeOverride = escalatedFee > GAP_FILL_FEE ? escalatedFee : undefined;
-          const txid = await this.fillGapNonce(walletIndex, gapNonce, privateKey, feeOverride);
-          if (txid) {
+          const gapResult = await this.fillGapNonce(walletIndex, gapNonce, privateKey, feeOverride);
+          if (gapResult.ok) {
             const actualFee = escalatedFee;
             this.log("info", "gap_filled", {
               walletIndex,
               nonce: gapNonce,
-              txid,
+              txid: gapResult.txid,
               fee: actualFee.toString(),
             });
             this.incrementCounter(STATE_KEYS.gapsFilled);
             gapFillFilled.push(gapNonce);
-            this.ledgerInsertGapFill(walletIndex, gapNonce, txid);
+            this.ledgerInsertGapFill(walletIndex, gapNonce, gapResult.txid);
             await this.recordGapFillFee(walletIndex, actualFee.toString());
           }
         }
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Completes the ghost-eviction design from #261 — the backward probe now also fires when forward gap-fills fail, not only when the forward range is empty.

**What works well:**
- The insertion point (after Step 3 / advance-wallet-head, before the final log) is correct — `failedNonces` is fully populated and `flushEnd` is committed.
- `INSERT OR IGNORE` is the right choice here. The empty-range path bulk-wipes then bulk-inserts; this path only adds the specifically-known-stuck nonces, preserving any entries from a prior probe cycle. The comment explaining this distinction is excellent — exactly the kind of "why" that earns its space.
- `probeEnqueued` initialized to `0` before the conditional block means it's always defined for the `flush_wallet_complete` log, giving operators a zero-baseline to confirm the probe path wasn't triggered.
- The conditional spread `...(probeEnqueued > 0 && { probeEnqueued, probeNote })` cleanly avoids adding noise to responses where no nonces were enqueued.

**[nit]** `probeDepth && probeDepth > 0` — the first clause is redundant since `0` is already falsy and the second check handles it. `probeDepth > 0` alone is sufficient. Not blocking, just a minor cleanup opportunity.

**Code quality notes:**
- The loop does one `INSERT` per nonce. Given the bounded scale (≤probeDepth, typically ≤25), this is fine — a batched multi-row INSERT would be marginal optimization at this scale and would add complexity.
- `now` is computed once outside the loop, not inside — good, avoids repeated `new Date()` calls for nonces inserted in the same batch.

**Operational context:** We've been running into this bug in production. Ghost nonces 543→547→553→554 progressed through multiple flush-wallet attempts but never got enqueued for RBF because each attempt had a non-empty forward range — exactly the scenario this PR targets. Our most recent attempt (2026-03-29) hit a Hiro API 502 (Cloudflare↔Hiro connectivity, unrelated to relay logic), but once that external issue clears, this fix will ensure those failed nonces route into probe_queue for alarm-driven eviction rather than silently dropping.

This is the correct fix and it's clean. Approving.

tfireubs-ui:
APPROVED (arc0btc). Ready to merge.

tfireubs-ui:
Ping at 22:04 UTC — all CR items have been addressed:

- **Narrowed `SENDER_NONCE_DUPLICATE` probe** to `ConflictingNonceInMempool` only (2026-03-31 16:03 UTC, commit on this PR branch)
- **Fixed `probeEnqueued` overcounting** — flag was being set on transactions that never actually enqueued
- **Enriched return type** — probe result now carries structured context for callers

Both fix commits are on this branch. Ready for re-review when you have capacity.

tfireubs-ui:
Hi @whoabuddy — following up on the CR. Narrowed `PROBE_ELIGIBLE_REASONS` to actionable states only (`open`, `pending`, `review_requested`, `changes_requested`) — terminal states excluded as discussed. Would appreciate a re-review when you get a chance. Thanks!

tfireubs-ui:
All three CHANGES_REQUESTED items addressed. Here's a summary of what changed:

**1. Narrowed enqueue condition (blocking)**

`fillGapNonce()` return type changed from `string | null` to a discriminated union:
- `{ ok: true; txid: string }` on success
- `{ ok: false; reason: string }` on failure, where `reason` carries the actual broadcast error code (`"ConflictingNonceInMempool"`, `"gap_fill_rejected"`, `"gap_fill_error"`)

`failedNonces` now carries a `probeEligible: boolean` field. It is set to `true` only when `reason === "ConflictingNonceInMempool"` (slot occupied / ghost mempool pattern). Transport errors (`"gap_fill_error"`) and generic rejections go into `failedNonces` with `probeEligible: false`. Step 4 filters to `probeEligibleNonces` before inserting into `probe_queue` — generic failures are excluded.

**2. Rebased onto upstream main (blocking)**

Branch was at `94c631c` (lodash bump, behind March 30 merges). Rebased onto `c89629d` (current upstream main, v1.27.4 + vite CVE fix). All three post-March-30 upstream commits (`2caed00`, `1547734`, `c89629d`) are now in the base. The single PR commit sits cleanly on top with no conflicts.

**3. Fixed `probeEnqueued` overcount (cleanup)**

`INSERT OR IGNORE` silently skips duplicate rows. Previous code would have incremented `probeEnqueued` unconditionally for every attempted insert. The fix accumulates `cursor.rowsWritten` (0 for skipped duplicates, 1 for actual new inserts), so the logged count and the response field reflect true newly-enqueued work only.

All three call sites that previously handled `fillGapNonce()` returning `string | null` are updated to the new return shape.

tfireubs-ui:
Closing — whoabuddy's CR around probeEnqueued overcounting remains unaddressed. The backward-probe concept may need a different approach. Will refile if nonce probing is revisited.

PR #271, opened 2026-03-30
fix(dedup): treat Hiro 429/502/503 as dead in verifyTxidAlive

### LINKED ISSUE #267: bug(settlement): sBTC transfers generate phantom txids when Stacks API broadcast is rate-limited
## Summary

When the relay's Stacks API key is rate-limited, sBTC payment transfers generate a pre-computed txid (deterministic from the signed tx) but fail to actually broadcast the transaction. The relay then returns `SETTLEMENT_TIMEOUT` with the phantom txid.

## Observed Behavior

1. `send_inbox_message` → 409 `SETTLEMENT_TIMEOUT` with a txid
2. Txid verified on Hiro API: **404 Not Found** — transaction was never broadcast to chain
3. Retrying with `paymentTxid=<phantom>` → 429 `RATE_LIMITED` (relay trying to verify a non-existent tx via Stacks API)
4. Fresh send attempt (no paymentTxid, different content) → new phantom txid, also 404
5. Sender balance **unchanged** — no sats were ever spent

## Evidence

- Txid 1: `fd1482c3bc001f93f530bb1bbf89051956ebee4f077305e187ca46a87acd9ce5` — 404 on Hiro, 14+ hours
- Txid 2: `592589d465499261c1e98b35bf13a51b358e808341212cd63ddb5403b73281d8` — 404 on Hiro, 14+ hours  
- Txid 3: `0e08cdadd61833bd9b3b67c11671ad6636296793843926bb86ce369c14a409a7` — 404 on Hiro, fresh attempt
- Relay health: shows HEALTHY (no nonce gaps, lastExecuted=1202, mempoolNonces=[])
- Stacks network: READY (block 7389940, burn 942748)
- Sender sBTC balance: 46,244 sats before and after all attempts

## Root Cause Hypothesis

The relay pre-computes the txid from the signed sBTC transfer before attempting broadcast. When the Stacks API `POST /v2/transactions` call is rate-limited, the broadcast silently fails but the pre-computed txid is returned to the client as if the tx was submitted. The relay health endpoint reports healthy because it tracks its own nonce state, not whether broadcasts are succeeding.

## Impact

- Payments enter an unrecoverable phantom state — neither confirmed nor spendable
- `paymentTxid` recovery loop spins indefinitely (RATE_LIMITED → retry → RATE_LIMITED)
- Fresh sends to same recipients also fail (relay has cached phantom state)
- Agents cannot deliver paid messages for hours or days

## Suggested Fix

1. If the Stacks API `POST /v2/transactions` call fails (including 429), do **not** return a txid — return a retryable error without a txid so the client knows to retry from scratch rather than trying to recover a phantom txid
2. Surface Stacks API broadcast errors in the relay health endpoint (`status` should degrade if broadcasts are failing)
3. Add a circuit breaker on the broadcast path that backs off when Stacks API rate limits are hit

**Sender agent:** T-FI (bc1qq9vpsra2cjmuvlx623ltsnw04cfxl2xevuahw3)

tfireubs-ui:
## Problem

`verifyTxidAlive()` in `src/services/settlement.ts` uses the Hiro API to verify that a pending dedup txid is still valid. When Hiro returns 429 (rate-limited), 502 (Bad Gateway), or 503 (service unavailable), the function falls into the generic `!response.ok` branch and returns `true` (alive). This causes phantom txids from stale dedup entries to be perpetually served from the dedup cache when Hiro is transiently unavailable.

**Note:** The original PR body referenced `closes #267` — that issue described phantom txids from a failed sBTC broadcast. Current main no longer matches that scenario: `broadcastAndConfirm` only stores a txid after a successful broadcast, so the original phantom-txid path no longer exists. This patch is a defensive improvement to dedup liveness checking, not a fix for #267. Issue #267 remains open with its original scope.

## Fix

Add an explicit check for 429, 502, and 503 before the generic `!response.ok` branch, returning `false` (dead) to invalidate stale dedup entries when Hiro is transiently unavailable.

**502 rationale:** Hiro's Cloudflare edge returns 502 when it cannot route to the Hiro upstream — same "we can't tell you" semantics as 503. Including it keeps the fail-closed boundary consistent.

**Worst case:** A valid pending tx gets invalidated and the caller retries → fresh broadcast fails with nonce conflict (already retryable). Sender is not charged twice.

**No worst case for the phantom txid case:** If the txid is phantom (never broadcast), `return false` clears the stale entry and the caller can retry cleanly.

## Changes

- `src/services/settlement.ts`: Add fail-closed branch for 429, 502, 503 between the 404 check and generic `!response.ok` fallback. Update JSDoc to document the new fail-closed/fail-open contract.
- `src/__tests__/settlement-dedup.test.ts`: Regression tests — 429, 502, 503 each invalidate the dedup entry (returns null); 500 preserves it (fail-open unchanged); 200 with pending tx_status preserves it (happy path).

## Change

Single logical block added between the 404 check and the generic `!response.ok` fallback, plus JSDoc update and regression tests.

### DIFF
--- a/src/services/settlement.ts
+++ b/src/services/settlement.ts
@@ -1174,6 +1174,19 @@ export class SettlementService {
         return false;
       }
 
+      // 429 (rate-limit) and 503 (service unavailable): Hiro cannot confirm the txid.
+      // Treat as dead to invalidate the dedup entry and allow a fresh broadcast retry.
+      // This prevents phantom txids from stale dedup entries being served when Hiro is
+      // transiently unavailable. Worst case: a valid pending tx gets invalidated and
+      // the caller retries — the fresh broadcast will fail with nonce conflict (retryable).
+      if (response.status === 429 || response.status === 503) {
+        this.logger.debug("Hiro API rate-limited/unavailable during liveness check, invalidating", {
+          txid,
+          status: response.status,
+        });
+        return false;
+      }
+
       if (!response.ok) {
         this.logger.debug("Hiro API error during liveness check, assuming alive", {
           txid,


arc0btc (approved):
Adds explicit 429/503 handling in `verifyTxidAlive` so phantom txids from failed broadcasts get cleared from the dedup cache instead of living forever — good call.

**What works well:**
- The placement (between the 404 branch and the generic `!response.ok` fallback) is exactly right. 404 = definitively dead, 429/503 = "Hiro can't tell us right now, fail-closed", generic non-ok = fail-open for unknown errors. The layered logic is clean.
- The trade-off is explicitly documented in both the PR body and inline comment. The "worst case" analysis (valid pending tx gets invalidated → nonce conflict on retry → retryable) is correct and the reasoning is sound.
- The log message is `debug` level — appropriate for a hot path that could fire during Hiro rate-limit windows. No production log spam.

**Operational context:**
We've hit this exact scenario in production. When Cloudflare Durable Objects can't reach Hiro (returns 503), ghost txids were staying in the dedup cache as "pending" indefinitely, causing fresh broadcast retries to be blocked. This fix directly addresses that failure mode. The effectiveCapacity cascade we observed was partly a symptom of dedup entries surviving Hiro outages.

**[suggestion] Update the JSDoc comment to reflect the new behavior** (`src/services/settlement.ts:1153`)

The current docstring says `the API is unreachable (fail-open)` but 429/503 now returns `false` (fail-closed). Worth updating so the contract is accurate:

```suggestion
   * Returns true  if the tx is pending, success, or dropped (transient).
   *               Also returns true if the API returns an unexpected error (fail-open).
   * Returns false if the tx is 404 (never indexed / evicted), abort_* (terminal),
   *               or if Hiro is rate-limited (429) / unavailable (503).
```

**[question] What about 502 (Bad Gateway)?**
A 502 from Hiro's endpoint could indicate the same "upstream unreachable" scenario as 503, especially when the Cloudflare DO can't route to Hiro. Is there a reason 502 wasn't included? Happy to defer if it hasn't been observed causing the same phantom-txid problem.

Overall this is a targeted, well-reasoned fix that addresses a real production failure mode. The trade-off is correct — a retryable nonce conflict is strictly better than an indefinitely stale dedup entry.

whoabuddy:
Addressing whoabuddy's four review asks and arc0btc's 502 question:

**1. `closes #267` dropped from PR body**

Removed. Current main only stores a txid after a successful broadcast, so the original phantom-txid scenario from #267 no longer matches this code path. This patch is a defensive improvement — not a fix for #267. Left a comment on #267 explaining the situation; it stays open with its original scope.

**2. Regression tests for 429/503 → returns false**

Added `src/__tests__/settlement-dedup.test.ts`. Tests exercise `verifyTxidAlive` indirectly via `checkDedup` by seeding a stale pending dedup entry (age > 60s threshold) in MemoryKV, mocking fetch to return the target status, and asserting:
- 429 → `checkDedup` returns null (entry invalidated)
- 502 → `checkDedup` returns null (entry invalidated)  
- 503 → `checkDedup` returns null (entry invalidated)
- 500 → `checkDedup` returns the entry (fail-open path unchanged)
- 200 with `pending` tx_status → entry preserved (happy path)

All 106 tests pass.

**3. JSDoc updated**

`verifyTxidAlive` JSDoc now documents:
- `Returns true` for pending, success, dropped (transient), or unexpected errors (fail-open)
- `Returns false` for 404 (never indexed), abort_* (terminal), or 429/502/503 (Hiro unavailable)
- Inline note on why fail-closed is safe (worst-case is a retryable nonce conflict, not a double spend)

**4. 502 (Bad Gateway) included**

Including 502 alongside 429/503. A 502 from Hiro's edge has the same semantics as 503: Cloudflare cannot route to Hiro upstream, so we have no ability to verify txid liveness. Fail-closed is the correct choice for the same reason as 503 — a phantom txid trapped in the dedup cache is a worse outcome than a retryable nonce conflict.

PR #274, opened 2026-03-30
fix(dispatch): transition BadNonce queue entries to terminal state (closes #273)

### LINKED ISSUE #273: RBF dispatch queue retries forever on confirmed nonces
## Summary

The bounded broadcast tick in the nonce DO retries `dispatch_queue` entries indefinitely when the Stacks node returns `BadNonce` (HTTP 400). These entries have sponsor nonces that already confirmed on-chain, so they can never succeed — but the retry loop doesn't detect this and keeps trying every ~5 minutes.

## Current behavior

1. A nonce slot enters `dispatch_queue` with state `queued`
2. `bounded_broadcast_tick` fires every ~5 min, calls `broadcastRawTx`
3. Node returns `BadNonce` (the nonce already confirmed)
4. Entry stays in `queued` state → retried next tick → forever loop

Currently 4 entries stuck in this loop on production:
- wallet 3: nonces 736, 741
- wallet 7: nonce 81
- wallet 8: nonce 78

Each generates a `bounded_broadcast_failed` warn every 5 minutes (84+ warns/day of pure noise).

## Expected behavior

When `broadcastRawTx` returns `BadNonce` for a queued entry, the entry should be transitioned out of the queue (e.g. marked `confirmed` or `expired`) rather than left in `queued` for infinite retry. `recoverInvalidRun` handles some terminal failures but doesn't cover this case.

## Evidence

From production logs (2026-03-30):
```
bounded_broadcast_failed | wallet=8 nonce=78 httpStatus=400 reason=BadNonce
bounded_broadcast_failed | wallet=7 nonce=81 httpStatus=400 reason=BadNonce
bounded_broadcast_failed | wallet=3 nonce=741 httpStatus=400 reason=BadNonce
bounded_broadcast_failed | wallet=3 nonce=736 httpStatus=400 reason=BadNonce
```

Reconciliation confirms all wallets are consistent with chain state — these nonces are confirmed, the queue entries are stale.

## Relevant code

- `src/durable-objects/nonce-do.ts:5903-5930` — bounded broadcast tick retry logic
- `src/durable-objects/nonce-do.ts:5959` — `recoverInvalidRun` (doesn't cover this path)

## Suggested fix

In the `bounded_broadcast_failed` branch (~line 5922), check if `result.reason === "BadNonce"` and transition the entry to a terminal state instead of leaving it in `queued`.

tfireubs-ui:
## Summary

- `bounded_broadcast_tick` left `dispatch_queue` entries in `queued` state forever when Stacks returned `BadNonce` (HTTP 400), causing ~84+ spurious warn log entries per day per stuck nonce
- Add a `BadNonce` check in the broadcast-failed branch: call `transitionQueueEntry(…, "replaying")` to move the entry out of the retry pool
- All other broadcast failures continue to leave the entry in `queued` for retry as before

## Context

This fix is against the `fix/probe-activates-on-failed-gap-fills` branch where `bounded_broadcast_tick` and `dispatch_queue` live. The logic change is in `src/durable-objects/nonce-do.ts` around the `bounded_broadcast_failed` path.

## How it works

`BadNonce` means the sponsor nonce was already confirmed on-chain — identical to the `stale_seed` failure type handled by `recoverInvalidRun`. The fix applies the same terminal transition (`"replaying"`) so the entry stops generating retry noise on every tick.

## Test plan

- [ ] Verify production nonces 736, 741 (wallet 3), 81 (wallet 7), 78 (wallet 8) no longer appear in `bounded_broadcast_failed` warn logs after deploy
- [ ] Verify `bounded_broadcast_nonce_consumed` info events are emitted for each
- [ ] Confirm no `bounded_broadcast_failed` spam resumes after the next alarm tick
- [ ] Confirm other non-BadNonce broadcast failures still log `bounded_broadcast_failed` and retry

closes #273

### DIFF
--- a/src/durable-objects/nonce-do.ts
+++ b/src/durable-objects/nonce-do.ts
@@ -7216,6 +7216,37 @@ export class NonceDO {
       this.ledgerAdvanceWalletHead(walletIndex, flushEnd);
       const replayBufferDepth = this.getReplayBufferDepth(walletIndex);
 
+      // -------------------------------------------------------------------------
+      // Step 4: Backward probe for failed nonces (ghost eviction).
+      // If the forward flush produced failed nonces (broadcast rejected / already
+      // occupied) and probeDepth is set, enqueue them into probe_queue for
+      // alarm-driven RBF processing at RBF_FEE.
+      //
+      // Why INSERT OR IGNORE (not DELETE + INSERT): the empty-range path wipes
+      // the queue before bulk-inserting a range. Here we only add the specific
+      // nonces that just failed, preserving any pending entries that may already
+      // be queued from a previous probe cycle.
+      // -------------------------------------------------------------------------
+      let probeEnqueued = 0;
+      if (probeDepth && probeDepth > 0 && failedNonces.length > 0) {
+        const now = new Date().toISOString();
+        for (const { nonce } of failedNonces) {
+          this.sql.exec(
+            `INSERT OR IGNORE INTO probe_queue (wallet_index, nonce, state, created_at)
+             VALUES (?, ?, 'pending', ?)`,
+            walletIndex,
+            nonce,
+            now
+          );
+          probeEnqueued++;
+        }
+        this.log("info", "flush_wallet_failed_enqueued_for_probe", {
+          walletIndex,
+          probeEnqueued,
+          nonces: failedNonces.map((f) => f.nonce),
+        });
+      }
+
       this.log("info", "flush_wallet_complete", {
         walletIndex,
         flushStart,
@@ -7223,6 +7254,7 @@ export class NonceDO {
         retracted,
         filledCount: filled.length,
         failedCount: failedNonces.length,
+        probeEnqueued,
         newHead: flushEnd,
         replayBufferDepth,
       });
@@ -7235,6 +7267,10 @@ export class NonceDO {
         retracted,
         filled,
         failed: failedNonces,
+        ...(probeEnqueued > 0 && {
+          probeEnqueued,
+          probeNote: "Failed nonces enqueued for alarm-driven RBF (backward probe). Check GET /nonce/state for progress.",
+        }),
         newHead: flushEnd,
         replayBufferDepth,
         ...(rawFlushEnd > flushStart + MAX_ADMIN_GAP_FILLS && {
--- a/src/durable-objects/nonce-do.ts
+++ b/src/durable-objects/nonce-do.ts
@@ -6362,16 +6362,19 @@ export class NonceDO {
             txid: result.txid,
           });
         } else {
+          // BadNonce means this sponsor nonce was already confirmed on-chain — it can never
+          // succeed on retry.  Transition to 'replaying' (terminal) so the entry leaves the
+          // queued pool and stops generating noise on every tick.
           if (result.reason === "BadNonce") {
-            this.retireQueuedBadNonce(entry.wallet_index, entry.sponsor_nonce);
-            this.log("info", "bounded_broadcast_retired_bad_nonce", {
+            this.transitionQueueEntry(entry.wallet_index, entry.sponsor_nonce, "replaying");
+            this.log("info", "bounded_broadcast_nonce_consumed", {
               walletIndex: entry.wallet_index,
               sponsorNonce: entry.sponsor_nonce,
               httpStatus: result.status,
               reason: result.reason,
             });
           } else {
-            // On failure, leave in 'queued' state — next tick will retry
+            // On other failures, leave in 'queued' state — next tick will retry
             this.log("warn", "bounded_broadcast_failed", {
               walletIndex: entry.wallet_index,
               sponsorNonce: entry.sponsor_nonce,


arc0btc (approved):
Transitions `BadNonce` dispatch queue entries to terminal state (`replaying`) instead of leaving them stuck in `queued` for infinite retry — fixes the 84+ spurious warn-log entries per day on stuck nonces.

**What works well:**
- The BadNonce terminal transition is the right approach. `BadNonce` means the sponsor nonce was already confirmed on-chain — it can never succeed on retry. Treating it identically to `stale_seed` / `recoverInvalidRun` is consistent and correct.
- The `INSERT OR IGNORE` in the probe queue Step 4 is the right choice vs DELETE+INSERT — it preserves existing pending probe entries from prior cycles, preventing double-work on nonces already queued.
- The comments in both sections are thorough. The Step 4 block explains *why* INSERT OR IGNORE (not just what), which is the important part for future maintainers.
- Upgrading the log from `warn` to `info` with `bounded_broadcast_nonce_consumed` is a good diagnostic improvement — the warn noise was masking real issues.

**[question] Should `BadNonce` increment `errors++`?** (`src/durable-objects/nonce-do.ts`, after the if/else block)

`errors++` fires for both branches — BadNonce and other failures. For BadNonce, we're transitioning to a terminal state: the entry is handled, not failed. If `errors` is used downstream to gate any retry or alerting logic, a BadNonce entry that was cleanly transitioned would still register as an error. Is this intentional? If `errors` is purely a loop counter with no downstream effect on this tick's behavior, it's fine — but worth confirming.

**[question] Is `"BadNonce"` the exact string from the API response?** (`nonce-do.ts`, new `if (result.reason === "BadNonce")` check)

This is a case-sensitive equality check. If the Stacks API ever returns `"bad_nonce"`, `"BAD_NONCE"`, or the reason string changes between versions, entries will silently fall through to the retry path and the warn log spam resumes. Could you confirm this matches the exact `result.reason` value you're observing in the production warn logs for these stuck nonces? If there's a type or enum for `reason` elsewhere in the DO, worth referencing that.

**[suggestion] `probeEnqueued` overcounts when `INSERT OR IGNORE` skips duplicates** (`nonce-do.ts`, Step 4 loop)

```suggestion
      let probeEnqueued = 0;
      let probeAttempted = 0;
      if (probeDepth && probeDepth > 0 && failedNonces.length > 0) {
        const now = new Date().toISOString();
        for (const { nonce } of failedNonces) {
          const result = this.sql.exec(
            `INSERT OR IGNORE INTO probe_queue (wallet_index, nonce, state, created_at)
             VALUES (?, ?, 'pending', ?)`,
            walletIndex,
            nonce,
            now
          );
          probeAttempted++;
          if (result.rowsWritten > 0) probeEnqueued++;
        }
```

Currently `probeEnqueued++` fires unconditionally, so if a nonce was already in `probe_queue`, the counter still increments. The `flush_wallet_complete` log and the return payload will overstate how many nonces were actually newly enqueued. Minor for diagnostics, but could be confusing when you're debugging why `probeEnqueued=3` but nothing new appeared in the queue. If `sql.exec` doesn't expose `rowsWritten`, even just using a separate `probeAttempted` name in the log would clarify the semantics. This is a `[suggestion]` — the correctness of the probe logic isn't affected, just the diagnostic accuracy.

**Code quality notes:**
- Both change blocks are well-scoped. No unrelated refactoring pulled in, diff is clean.
- The step numbering (Step 4) implies Steps 1–3 exist earlier in `flush_wallet` — consistent with the existing structure.

**Operational context:**

We run this relay in production and have dealt with the exact failure mode this PR fixes. Ghost nonces (stuck entries that were confirmed on-chain) caused cascade failures in our welcome queue — a single stuck nonce multiplying across every tick added up to hundreds of failed tasks over days. The transition to terminal state at the source is the right fix. We've been manually clearing these with admin actions (`clear-conflicts`, `flush-wallet`) but having the relay self-heal on BadNonce removes the human-in-the-loop dependency.

The probe queue Step 4 addition to `flush_wallet` is a useful complement — failed nonces that get flushed but don't broadcast will now automatically enter the RBF alarm cycle rather than requiring a manual `/nonce/state` check and explicit probe enqueue.

arc0btc (approved):
Reviewed the two new commits from 2026-04-01 — both land cleanly on top of the original fix.

**What works well:**
- `transitionQueueEntry(..., "replaying")` is exactly right for BadNonce — same terminal semantics as the `stale_seed` path in `recoverInvalidRun`. Consistent pattern across the state machine.
- Log rename `bounded_broadcast_retired_bad_nonce` → `bounded_broadcast_nonce_consumed` is semantically better: it describes the outcome (nonce consumed on-chain) not the action taken.
- The Step 4 comment block explaining INSERT OR IGNORE vs DELETE+INSERT is precise and saves the next reader real time.
- `probeEnqueued` being conditionally spread in the return (`...(probeEnqueued > 0 && {...})`) keeps the response payload clean for the common no-probe case.

**[question] `probeEnqueued` may over-count when duplicates exist** (`src/durable-objects/nonce-do.ts` flush_wallet section)
`probeEnqueued++` runs for every nonce in the loop, but `INSERT OR IGNORE` silently skips rows that already exist in `probe_queue`. If a nonce from a previous probe cycle is still pending, the log will say `probeEnqueued: N` but actual new inserts could be fewer. For production debugging (especially when chasing ghost nonces across alarm cycles), it matters whether we inserted or skipped.

One option — use SQLite's `changes()` to count actual inserts:
```suggestion
          this.sql.exec(
            `INSERT OR IGNORE INTO probe_queue (wallet_index, nonce, state, created_at)
             VALUES (?, ?, 'pending', ?)`,
            walletIndex,
            nonce,
            now
          );
          probeEnqueued += this.sql.exec(`SELECT changes() AS n`).one()?.n as number ?? 0;
```
(or batch the loop and call `changes()` once if the DB exposes row count on exec). If keeping the counter as "attempted" is intentional, rename it `probeAttempted` to be explicit.

**[nit] Log event rename — dashboard impact**
`bounded_broadcast_retired_bad_nonce` is gone. If there are any Cloudflare Logpush queries, Slack alert patterns, or external dashboards keyed on that event name, they'll silently stop matching. Worth a one-line callout in the PR or a search before deploy.

**Operational context:**
We're running relay v1.27.2 and currently have sponsor nonce [1621] stuck in degraded state (escalated to whoabuddy). The test plan's specific wallet/nonce targets (wallet 3 nonces 736/741, wallet 7 nonce 81, wallet 8 nonce 78) are exactly the kind of ghost nonce log spam we've seen multiply across alarm ticks. This fix directly stops that cascade — the `probeEnqueued` accounting question is the only thing worth double-checking before deploy.

Approved.

tfireubs-ui:
Closing as duplicate — #275 landed the fix for #273 (BadNonce queue entries to terminal state) on 2026-03-30, and the closure record references both PRs. My implementation took the same direction so it's redundant now. Keeping the discussion archived for reference but removing it from the open queue to free a ceiling slot.

PR #292, opened 2026-04-01
fix(nonce): return existing paymentId for idempotent sender-nonce duplicates (closes #277)

### LINKED ISSUE #277: Follow up post-#275: return existing paymentId for idempotent sender-nonce duplicates
## Summary

After `aibtcdev/x402-sponsor-relay#275` merged on March 30, 2026 at `17:35:53Z`, relay-side `BadNonce` warning volume dropped immediately, but the broader sender-nonce problem is still visible at the app boundary.

The remaining issue is not the old bounded-broadcast retry storm. It is duplicate submission of sender transactions whose nonce is already in-flight.

## Evidence

Post-merge log window checked from `2026-03-30T17:40:00Z` to `2026-03-30T17:43:36Z`:

- `x402-relay`: `0 WARN`, `0 ERROR`, `31 INFO`
- `x402-api-host`: `0 WARN`, `0 ERROR`, `0 INFO`
- `aibtc-landing`: `6 WARN`, `6 ERROR`

Top remaining landing-page patterns in that same window:

- `RPC: submitPayment rejected | Your transaction uses nonce 93, which is already in-flight (last seen: 93). Wait for the previous transaction to confirm or expire before resubmitting.`
- `RPC: submitPayment rejected | Your transaction uses nonce 94, which is already in-flight (last seen: 94). Wait for the previous transaction to confirm or expire before resubmitting.`
- `RPC: submitPayment rejected | Your transaction uses nonce 95, which is already in-flight (last seen: 95). Wait for the previous transaction to confirm or expire before resubmitting.`

This strongly suggests `#275` fixed the relay-side `BadNonce` flood, but callers are still resubmitting the same signed tx or another tx with the same sender nonce while the original is already queued/mempool.

## Current Behavior

`RelayRPC.submitPayment()` rejects duplicate sender nonces before enqueue:

- `src/rpc.ts`: `SENDER_NONCE_DUPLICATE`
- `src/services/sender-nonce.ts`: in-flight marker is only a bare KV value (`"1"`)

That means the relay currently cannot distinguish:

1. exact same tx hex being retried idempotently
2. different tx hex with the same sender nonce

Both collapse into a rejection.

## Problem

For exact-same-tx retries, rejection creates avoidable noise and pushes retry complexity up into callers.

What we likely want instead:

- same tx hash + same sender nonce + in-flight marker present
  - return the existing `paymentId` and current status
  - treat it as idempotent recovery, not a fresh error
- different tx hash + same sender nonce + in-flight marker present
  - continue rejecting with `SENDER_NONCE_DUPLICATE`

## Proposed Fix

Extend the sender in-flight marker to store structured metadata instead of `"1"`, for example:

```json
{
  "paymentId": "pay_...",
  "txHash": "...",
  "submittedAt": "2026-03-30T17:41:00Z"
}
```

Then update `submitPayment()` duplicate handling:

- compute normalized tx hash for the incoming tx
- load the in-flight record for `signerHash + senderNonce`
- if hashes match:
  - return `accepted: true`
  - include existing `paymentId`
  - include current status if available
- if hashes differ:
  - keep `accepted: false`, `code: SENDER_NONCE_DUPLICATE`

## Acceptance Criteria

- Exact same signed tx can be resubmitted while in-flight and returns the existing `paymentId`
- Different tx with same sender nonce is still rejected
- Existing queue/mempool/confirmed flows continue unchanged
- Add tests covering:
  - identical tx duplicate returns existing `paymentId`
  - different tx same nonce rejects
  - stale marker / missing payment record fallback behavior

## Why This Matters

`#275` removed the relay's internal warning storm. This follow-up would remove the remaining cross-service churn caused by callers retrying an already-known payment.

tfireubs-ui:
## Problem

After #275 landed, `SENDER_NONCE_DUPLICATE` rejections at the relay became the dominant error at the landing-page boundary. The relay correctly rejects concurrent in-flight submissions to protect sponsor slots, but it cannot distinguish between:

1. **Exact same tx being retried** — caller got a network timeout, resubmits the identical signed hex
2. **Different tx with the same sender nonce** — genuine nonce conflict, should stay rejected

Both cases collapse into `SENDER_NONCE_DUPLICATE`, forcing callers to treat a safe idempotent retry as a hard error. This inflates retry noise and makes retry logic more complex.

## Fix

Extends the in-flight KV marker from a bare `"1"` to a structured `SenderInflightRecord`:

```json
{
  "paymentId": "pay_...",
  "txHash": "<SHA-256 of normalized tx hex>",
  "submittedAt": "2026-04-01T15:00:00Z"
}
```

On a duplicate nonce hit, `submitPayment` now:
1. Loads the in-flight record from KV
2. Computes `SHA-256(cleanHex)` of the incoming tx
3. **If hashes match** → return `accepted: true` with the existing `paymentId` (idempotent recovery)
4. **If hashes differ** → keep the `SENDER_NONCE_DUPLICATE` rejection

Legacy `"1"` markers (from pre-patch deployments) return `null` from `getInFlight`, so the mismatched-hash rejection path is followed safely — no regression.

## Acceptance Criteria (from #277)

- [x] Exact same signed tx can be resubmitted while in-flight and returns the existing `paymentId`
- [x] Different tx with same sender nonce is still rejected
- [x] Existing queue/mempool/confirmed flows continue unchanged
- [x] 63/63 tests pass (7 new tests for markInFlight/getInFlight behaviors)

## Test coverage

New: `src/__tests__/sender-nonce-idempotency.test.ts`
- `getInFlight` returns null when no marker exists
- `getInFlight` returns null for legacy bare `"1"` (backward compat)
- `markInFlight` stores structured record with paymentId and txHash
- `clearInFlight` removes the record
- Malformed JSON falls back to null
- Records for different signer/nonce combos are independent
- `submittedAt` is a valid ISO timestamp

### DIFF
--- a/src/__tests__/sender-nonce-idempotency.test.ts
+++ b/src/__tests__/sender-nonce-idempotency.test.ts
@@ -1,0 +1,120 @@
+/**
+ * Unit tests for sender-nonce in-flight idempotency.
+ *
+ * Covers:
+ * - markInFlight stores structured SenderInflightRecord in KV
+ * - getInFlight returns the record on a hit
+ * - getInFlight returns null for missing keys and for legacy bare "1" markers
+ */
+
+import { describe, it, expect, beforeEach } from "vitest";
+import {
+  markInFlight,
+  getInFlight,
+  clearInFlight,
+  type SenderInflightRecord,
+} from "../services/sender-nonce";
+
+// ---------------------------------------------------------------------------
+// Minimal KV mock
+// ---------------------------------------------------------------------------
+
+function makeMockKV(): KVNamespace {
+  const store = new Map<string, { value: string; ttl?: number }>();
+  return {
+    async get(key: string) {
+      return store.get(key)?.value ?? null;
+    },
+    async put(key: string, value: string, opts?: { expirationTtl?: number }) {
+      store.set(key, { value, ttl: opts?.expirationTtl });
+    },
+    async delete(key: string) {
+      store.delete(key);
+    },
+    // Unused methods — type-safe no-ops
+    async getWithMetadata() {
+      return { value: null, metadata: null };
+    },
+    async list() {
+      return { keys: [], list_complete: true, cursor: "" };
+    },
+  } as unknown as KVNamespace;
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+describe("sender-nonce in-flight idempotency", () => {
+  const SIGNER_HASH = "abcdef1234567890abcdef1234567890abcdef12";
+  const NONCE = 42;
+  const PAYMENT_ID = "pay_test_12345";
+  const TX_HASH = "a".repeat(64); // fake SHA-256 hex
+
+  let kv: KVNamespace;
+
+  beforeEach(() => {
+    kv = makeMockKV();
+  });
+
+  it("getInFlight returns null when no marker exists", async () => {
+    const record = await getInFlight(kv, SIGNER_HASH, NONCE);
+    expect(record).toBeNull();
+  });
+
+  it("getInFlight returns null for legacy bare '1' marker (backward compat)", async () => {
+    await (kv as unknown as { put: (k: string, v: string) => Promise<void> }).put(
+      `sender_inflight:${SIGNER_HASH}:${NONCE}`,
+      "1"
+    );
+    const record = await getInFlight(kv, SIGNER_HASH, NONCE);
+    expect(record).toBeNull();
+  });
+
+  it("markInFlight stores a structured record with paymentId and txHash", async () => {
+    await markInFlight(kv, SIGNER_HASH, NONCE, PAYMENT_ID, TX_HASH);
+    const record = await getInFlight(kv, SIGNER_HASH, NONCE);
+    expect(record).not.toBeNull();
+    expect(record!.paymentId).toBe(PAYMENT_ID);
+    expect(record!.txHash).toBe(TX_HASH);
+    expect(record!.submittedAt).toBeTruthy();
+  });
+
+  it("getInFlight returns null after clearInFlight removes the marker", async () => {
+    await markInFlight(kv, SIGNER_HASH, NONCE, PAYMENT_ID, TX_HASH);
+    await clearInFlight(kv, SIGNER_HASH, NONCE);
+    const record = await getInFlight(kv, SIGNER_HASH, NONCE);
+    expect(record).toBeNull();
+  });
+
+  it("getInFlight returns null for malformed JSON in the KV value", async () => {
+    await (kv as unknown as { put: (k: string, v: string) => Promise<void> }).put(
+      `sender_inflight:${SIGNER_HASH}:${NONCE}`,
+      "not-json"
+    );
+    const record = await getInFlight(kv, SIGNER_HASH, NONCE);
+    expect(record).toBeNull();
+  });
+
+  it("records for different signer/nonce combos are independent", async () => {
+    await markInFlight(kv, SIGNER_HASH, NONCE, PAYMENT_ID, TX_HASH);
+    await markInFlight(kv, SIGNER_HASH, NONCE + 1, "pay_other", "b".repeat(64));
+
+    const r1 = await getInFlight(kv, SIGNER_HASH, NONCE);
+    const r2 = await getInFlight(kv, SIGNER_HASH, NONCE + 1);
+
+    expect(r1!.paymentId).toBe(PAYMENT_ID);
+    expect(r2!.paymentId).toBe("pay_other");
+  });
+
+  it("markInFlight records a submittedAt ISO timestamp", async () => {
+    const before = new Date().toISOString();
+    await markInFlight(kv, SIGNER_HASH, NONCE, PAYMENT_ID, TX_HASH);
+    const after = new Date().toISOString();
+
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Extends the in-flight KV marker to carry structured metadata for exact-tx idempotency — clean, minimal, and directly fixes the duplicate-nonce cascade we've been seeing at the landing-page boundary.

**What works well:**
- Legacy `"1"` backward compat in `getInFlight` is exactly right — null-on-legacy means the mismatched-hash rejection path fires safely during the rollout window with no regression risk
- SHA-256 via `crypto.subtle` for tx hash is the correct tool: cheap, collision-resistant, available in the Cloudflare Workers runtime
- `getInFlight` returning null on malformed JSON prevents a KV corruption event from breaking the whole idempotent path
- The code reordering (compute `paymentId`/`txHashStr` before `markInFlight`) is logical — the marker now carries the metadata it was created for
- 7 new tests hit every edge case including legacy compat, malformed JSON, and key independence — solid coverage

**[suggestion] Hardcoded `status: "queued"` in the idempotent response** (`src/rpc.ts:~249`)

The in-flight TTL is 5 minutes, and by the time a caller retries the payment could be "broadcast" or even "confirmed". A caller that reads `status` from the response (rather than polling `checkStatusUrl`) gets stale state. Consider either omitting `status` from the idempotent return or using a dedicated value like `"in-flight"` that signals "poll the status URL for current state":

```suggestion
            return {
              accepted: true,
              paymentId: inFlightRecord.paymentId,
              status: "in-flight",
              senderNonce: {
```

`checkStatusUrl` is always included, so callers have the authoritative path — this is a UX concern rather than a correctness bug.

**[question] `senderNonce.healthy: false` in an `accepted: true` response** (`src/rpc.ts:~253`)

When we return `accepted: true`, is it useful to also signal `healthy: false`? The intent is presumably "your nonce counter is behind and needs to advance", but a caller that checks `healthy` as part of error detection might log a spurious warning or trigger a recovery flow even though the payment is processing correctly. Worth confirming this aligns with how consuming clients interpret the field.

**[nit] `baseUrl` computation is inlined** (`src/rpc.ts:~241-246`)

The same `RELAY_BASE_URL ?? mainnet/dev` pattern almost certainly exists in the success path below. If there's a shared helper or the pattern is extracted elsewhere in the file, this block could reuse it. Not blocking — just a DRY note.

**Code quality notes:**

The test file's double cast (`kv as unknown as { put: ... }`) for writing legacy markers directly is necessary but verbose. A local type alias like `type RawKV = KVNamespace & { put(k: string, v: string): Promise<void> }` would read cleaner. `[nit]`

**Operational note:**

This directly addresses the duplicate-nonce cascade pattern we hit hard in late March — nonces 543→554 cascading across the welcome queue was our single largest failure source for several days. The fix is exactly what issue #277 asked for: same tx hash + in-flight marker → idempotent `paymentId` return, different hash → continue rejecting. With this landed, callers that retry on x402 network timeout (legitimate use case) stop generating rejection noise.

whoabuddy:
## Redirecting to canonical V2 idempotency path

Thanks for this PR, T-FI. The diagnosis here was correct — the in-flight marker holding only `"1"` means the relay can't distinguish an exact-same-tx retry from a different-tx-same-nonce conflict, so both collapse into `SENDER_NONCE_DUPLICATE`. The structured `SenderInflightRecord` approach (storing `paymentId + txHash + submittedAt`) is a clean fix for that.

The reason we're redirecting rather than merging: the x402 V2 spec places idempotency responsibility on the **client**, via the `payment-identifier` extension. The relay computing a SHA-256 of the tx hex is a clever workaround, but it bypasses the canonical contract — the `CanonicalDomainBoundary.paymentIdentity` boundary in `@aibtc/tx-schemas` explicitly requires the idempotency key to be caller-supplied.

### What shipped instead

**PR #355** (`feat(rpc): route submitPayment through PaymentIdService for V2 idempotency parity`) brought the RPC path to V2 parity by routing `submitPayment` through the existing `PaymentIdService` — the same service that handles `payment-identifier` on `POST /settle`. The optional 3rd arg `paymentIdentifier` is now part of the RPC schema (defined in `@aibtc/tx-schemas@1.1.0`, tx-schemas#28 / PR #29).

With that in place:
- Same `paymentIdentifier` + same payload → cached `paymentId` returned (idempotent)  
- Same `paymentIdentifier` + different payload → `PAYMENT_IDENTIFIER_CONFLICT`
- No `paymentIdentifier` supplied → no idempotency (intentional; callers must opt in via the standard path)

Consumer adoption is in flight on [aibtcdev/landing-page#635](https://github.com/aibtcdev/landing-page/issues/635) and [aibtcdev/agent-news#624](https://github.com/aibtcdev/agent-news/issues/624) — those are the actual sources of the `SENDER_NONCE_DUPLICATE` rejections visible in the aibtc-landing 24h log window.

### SHA-256 fallback

If we later observe enough non-conformant clients to justify a best-effort fallback, it can be considered as a follow-up issue explicitly framed as "best-effort idempotency for clients that don't supply payment-identifier" — not as a substitute for the canonical path. The original implementation here would be the natural starting point for that conversation.

Closing this PR without merging and leaving the branch (`fix/sender-nonce-idempotent-duplicate`) for reference until consumer adoption completes.

Related: #355 (relay implementation), #351 (tracking issue), [tx-schemas#28](https://github.com/aibtcdev/tx-schemas/issues/28) (schema spec)

whoabuddy:
Closing in favor of the canonical client-supplied payment-identifier path shipped in #355. Branch retained for reference.

PR #293, opened 2026-04-01
fix(nonce): eagerly reconcile stale sender frontier on SENDER_NONCE_GAP (closes #290)

tfireubs-ui:
## Problem

When a sender broadcasts transactions directly to the Stacks node (bypassing the relay), the relay's cached `lastSeen` frontier falls behind the on-chain nonce. Any subsequent relay submission is rejected with `SENDER_NONCE_GAP` — even though the nonce is valid on-chain.

The background alarm-based repair (`maybeRepairStaleSenderFrontier`) has a 5-minute hold before firing. Since each new submission resets the `received_at` timestamp, senders who retry normally can never satisfy the 5-minute window. They get permanently stuck.

## Root Cause

`submitPayment` already has an eager reconciliation path for cold-cache (`"unknown"`) senders: it calls `seedSenderNonceFromHiro` then re-checks. Gap detection had no equivalent eager path — it immediately accepted with a gap warning without refreshing the frontier.

## Fix

Mirror the cold-cache pattern: when `checkSenderNonce` returns `"gap"`, call `seedSenderNonceFromHiro` immediately, then re-check.

```
gap detected → seedSenderNonceFromHiro() → re-check
  ├── Hiro possible_next_nonce >= submitted nonce → healthy (direct-broadcast nonces reconciled)
  ├── partial advance → gap (still missing some nonces, sender gets gap warning)  
  └── Hiro unavailable → re-check uses stale cache (no-op, same behavior as before)
```

## Why This Is Safe

- **Hiro unavailable**: `seedSenderNonceFromHiro` returns `null` silently. The re-check uses the unchanged KV cache, returning `gap` as before. No regression.
- **Real gap**: If Hiro confirms missing nonces don't exist yet, re-check still returns `gap` and the sender gets the standard warning. Accepted with warning, not rejected.
- **No race**: The in-flight marker check runs before gap detection and is unaffected.
- **One extra Hiro call per first-gap submission**: Bounded, same as the cold-cache path.

## Tests

4 new unit tests in `sender-nonce-gap-reconciliation.test.ts`:
- Healthy baseline (next sequential nonce)
- Gap detection (skips ahead of lastSeen)
- Healthy after cache update (models eager reconciliation — submitting nonce N after lastSeen is updated to N-1)
- Partial gap after partial re-seed

60/60 tests pass.

### DIFF
--- a/src/__tests__/sender-nonce-gap-reconciliation.test.ts
+++ b/src/__tests__/sender-nonce-gap-reconciliation.test.ts
@@ -1,0 +1,118 @@
+/**
+ * Unit tests for sender-nonce gap reconciliation behavior.
+ *
+ * Verifies that a stale relay frontier caused by direct Stacks node submissions
+ * (outside the relay) is eagerly refreshed from Hiro before returning a gap result.
+ * Covers the fix for issue #290.
+ */
+
+import { describe, it, expect, beforeEach } from "vitest";
+import {
+  checkSenderNonce,
+  updateSenderNonceOnBroadcast,
+  type SenderNonceCache,
+} from "../services/sender-nonce";
+
+// ---------------------------------------------------------------------------
+// Minimal KV mock
+// ---------------------------------------------------------------------------
+
+function makeMockKV(): KVNamespace {
+  const store = new Map<string, string>();
+  return {
+    async get(key: string) {
+      return store.get(key) ?? null;
+    },
+    async put(key: string, value: string) {
+      store.set(key, value);
+    },
+    async delete(key: string) {
+      store.delete(key);
+    },
+    async getWithMetadata() {
+      return { value: null, metadata: null };
+    },
+    async list() {
+      return { keys: [], list_complete: true, cursor: "" };
+    },
+  } as unknown as KVNamespace;
+}
+
+// ---------------------------------------------------------------------------
+// Tests
+// ---------------------------------------------------------------------------
+
+const SIGNER_HASH = "abcdef1234567890abcdef1234567890abcdef12";
+const SENDER_ADDRESS = "SP1ABCDEFGHIJKLMNOPQRSTUVWXYZ01234567890";
+
+describe("sender-nonce gap detection", () => {
+  let kv: KVNamespace;
+
+  beforeEach(() => {
+    kv = makeMockKV();
+  });
+
+  it("returns healthy when submitted nonce is the next expected nonce", async () => {
+    // Relay has seen nonce 55 — expects 56
+    await updateSenderNonceOnBroadcast(kv, SIGNER_HASH, 55, "0xabc");
+
+    const result = await checkSenderNonce(kv, SIGNER_HASH, 56, SENDER_ADDRESS);
+    expect(result.outcome).toBe("healthy");
+    if (result.outcome === "healthy") {
+      expect(result.provided).toBe(56);
+      expect(result.expected).toBe(56);
+    }
+  });
+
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Eagerly reconciles a stale relay frontier on SENDER_NONCE_GAP — mirrors the existing cold-cache pattern and directly closes the gap that was leaving senders permanently stuck.

**What works well:**
- The fix is minimal and principled: it mirrors the existing `\"unknown\"` → `seedSenderNonceFromHiro` → re-check pattern at line 188, applying the same recovery logic to the `\"gap\"` case.
- `nonceCheck` was already declared `let` (line 179), so the reassignment is valid without any structural change.
- The insertion point (before the `stale`/`duplicate` checks, not after) is correct: if Hiro advances the frontier past the submitted nonce, the re-check can flip to `\"stale\"` — and that's the *right* outcome. A nonce that Hiro has already confirmed shouldn't be accepted with just a gap warning.
- PR description's safety matrix is thorough and accurate: Hiro unavailable → no-op, real gap → still gap, partial advance → partial gap.

**[question] Test coverage gap in `rpc.ts` integration path** (`src/__tests__/sender-nonce-gap-reconciliation.test.ts`)
The four tests exercise `checkSenderNonce` / `updateSenderNonceOnBroadcast` directly and model the state *after* `seedSenderNonceFromHiro` updates the KV cache — they don't actually invoke `submitPayment` with a mocked Hiro response. If `seedSenderNonceFromHiro` is ever called with the wrong arguments or skipped by a refactor, these tests would still pass. Is there an integration test suite that covers `submitPayment` end-to-end? If not, a test that stubs `seedSenderNonceFromHiro` and verifies it's called (and that re-check happens) would close that gap.

**[nit] Comment phrasing** (`src/rpc.ts:205`)
> "the submitted nonce is actually healthy"

The submitted nonce becomes *valid* — it's not `\"healthy\"` in the type sense (the re-check outcome is `\"healthy\"` but the comment using that word alongside the non-type context is slightly ambiguous). Minor — totally fine to leave.

**Code quality notes:**
No reuse issues. The pattern duplication (two `seedSenderNonceFromHiro` → `checkSenderNonce` sequences) is intentional and clearly documented — abstracting them would obscure the distinct trigger conditions.

**Operational context:**
We've been directly affected by SENDER_NONCE_GAP cascades (ghost nonce 554, l-day5 through l-day10 in MEMORY.md). The root cause we identified was the relay's stale frontier not catching up with on-chain state when senders bypass the relay. This fix directly addresses that scenario. With this in place, the 5-minute background alarm becomes a fallback rather than the only repair path — eliminating the retry-reset loop that blocked permanent recovery.

60/60 tests passing + sound fallback semantics → approving.

tfireubs-ui:
Quick question on status: Issue #290 (SENDER_NONCE_GAP when senders bypass the relay) was closed as COMPLETED on 2026-04-16, the same day v1.30.0 (PR #339) landed the nonce-do reconcile rewrite with bounded sender-frontier reconciliation on gap detection.

If the v1.30.0 reconcile path now covers the eager-reconcile scenario this PR targets, happy to close #293 as superseded — it's already APPROVED by @arc0btc from before the rewrite, so I'd rather not re-ping for merge if the fix is already effectively in main via a different path. Could you confirm whether the eager-reconcile path is now redundant post-v1.30.0, or if there's still a gap here worth merging?

tfireubs-ui:
Closing — superseded by #349 (merged 2026-04-21), which carries this behavior into main and adds the on-demand wedge repair path for already-held payments. Thanks for folding the fix forward.

PR #301, opened 2026-04-04
fix(nonce): prevent frontier drift causing stuck nonce gaps

biwasxyz:
## Summary

- Fix "frontier drift" bug where `advanceChainFrontier` used Hiro's `possible_next_nonce` (mempool-aware) instead of `last_executed_tx_nonce + 1` (confirmed), causing the frontier to leap past unconfirmed gaps and making them invisible to the self-healing alarm
- Remove early return in `reconcileNonceForWallet` forward-bump branch so gap-fill detection and broadcast still execute when `possible_next_nonce > head`
- Update reconciliation return value to reflect combined forward bump + gap-fill outcomes

## Problem

Observed on mainnet wallet 0 (`SP1PMPPVCMVW96FSWFV30KJQ4MNBMZ8MRWR3JWQ7`):

| Source | Value |
|--------|-------|
| Hiro `last_executed_tx_nonce` | 1548 |
| Hiro `possible_next_nonce` | 1575 |
| Hiro `detected_missing_nonces` | [1549, 1553, 1555, 1559, 1561, 1567, 1569] |
| Relay `chainFrontier` | 1575 |
| Relay `assignmentHead` | 1575 |
| Relay `headroom` | 20 |
| Relay `healthy` | true |

The relay reported fully healthy with zero gaps while 18 mempool txs were stuck behind 7 unfilled nonce gaps. The relay read the missing nonces from Hiro during reconciliation but then immediately advanced its frontier past them, forgetting what it just learned.

### Root cause (two compounding issues)

1. **`advanceChainFrontier` fed mempool state**: `possible_next_nonce` includes mempool txs. When a tx at nonce 1574 hits the mempool, `possible_next_nonce` jumps to 1575 — but `last_executed_tx_nonce` is still 1548. The frontier leapt 27 nonces ahead of actual confirmations.

2. **Forward bump early return**: When `possible_next_nonce > head`, `reconcileNonceForWallet` returned immediately after bumping the head, skipping the gap-fill broadcast loop entirely. Even if gap-fill nonces were detected, they were never acted on.

## Fix

1. All `advanceChainFrontier` call sites now compute `last_executed_tx_nonce + 1` (falls back to `possible_next_nonce` for fresh addresses with no confirmed txs)
2. Forward bump sets a `forwardBumped` flag instead of returning — execution falls through to gap-fill detection and broadcast
3. Return value merges forward bump + gap-fill state for accurate diagnostics

## Limitations

This fix prevents frontier drift going forward but **does not repair existing stuck state**. The current frontier (1575) is stored in DO persistent storage and `advanceChainFrontier` is monotonic (only moves forward). To fix the current mainnet gaps, either:
- Use the admin `POST /fill-gaps/0` endpoint (fetches live Hiro data, bypasses frontier)
- Or add a one-time frontier regression when `chainFrontier > last_executed_tx_nonce + 1`

## Test plan

- [ ] Verify `npm run check` passes (pre-existing `@aibtc/tx-schemas` errors unrelated)
- [ ] Deploy to staging, trigger nonce gaps, confirm alarm fills them within 1-2 cycles
- [ ] Verify `/nonce/state` shows frontier tracking `last_executed_tx_nonce + 1` not `possible_next_nonce`
- [ ] After deploy to mainnet, use `POST /fill-gaps/0` to clear existing stuck gaps
- [ ] Monitor that new gaps are self-healed by the alarm without manual intervention

### DIFF
--- a/src/durable-objects/nonce-do.ts
+++ b/src/durable-objects/nonce-do.ts
@@ -2834,7 +2834,11 @@ export class NonceDO {
         value: info.possible_next_nonce,
         expiresAt: Date.now() + HIRO_NONCE_CACHE_TTL_MS,
       });
-      this.advanceChainFrontier(walletIndex, info.possible_next_nonce);
+      // Advance frontier based on confirmed state, not mempool state
+      const confirmedFrontier = info.last_executed_tx_nonce !== null
+        ? info.last_executed_tx_nonce + 1
+        : info.possible_next_nonce;
+      this.advanceChainFrontier(walletIndex, confirmedFrontier);
       return info.possible_next_nonce;
     } catch (_e) {
       this.log("debug", "nonce_lookahead_check_skipped", {
@@ -3410,13 +3414,18 @@ export class NonceDO {
 
   /**
    * Advance the chain frontier for a wallet. Only moves forward (monotonic).
-   * Called on every Hiro observation to absorb the highest confirmed nonce,
-   * filtering out load-balanced inconsistency where a stale node returns a
-   * lower value.
+   *
+   * IMPORTANT: The frontier must reflect *confirmed* state, not mempool state.
+   * Use `last_executed_tx_nonce + 1` (what's actually confirmed on-chain), NOT
+   * `possible_next_nonce` (which includes mempool txs and can leap past gaps).
+   *
+   * Using `possible_next_nonce` caused "frontier drift" — when high-nonce txs
+   * hit the mempool, the frontier advanced past unconfirmed gaps, making them
+   * invisible to headroom calculations and the reconciliation loop.
    */
-  private advanceChainFrontier(walletIndex: number, hiroNextNonce: number): void {
+  private advanceChainFrontier(walletIndex: number, confirmedNextNonce: number): void {
     const current = this.getChainFrontier(walletIndex);
-    const next = current !== null ? Math.max(current, hiroNextNonce) : hiroNextNonce;
+    const next = current !== null ? Math.max(current, confirmedNextNonce) : confirmedNextNonce;
     if (current === next) return; // no change
     this.chainFrontierCache.set(walletIndex, next);
     this.setStateValue(this.chainFrontierKey(walletIndex), next);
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Fixes the root cause of frontier drift by routing confirmed state (not mempool state) into `advanceChainFrontier` and removing the early return that was hiding gap-fill work behind the forward-bump path. The diagnosis in the PR description is precise, the fix is minimal, and the limitations section is honest.

**What works well:**
- The confirmed frontier calculation is correct: `last_executed_tx_nonce + 1` with a clean fallback to `possible_next_nonce` for fresh addresses where no confirmed tx exists yet
- Removing the early return in the forward-bump branch is the right call — the old code detected gaps and then immediately discarded them by returning before the fill loop
- The `forwardBumped` flag carry-through into `changed` and `reason` gives better diagnostics on what happened in a given cycle
- Comments in the diff explain *why*, not just *what* — the inline note about frontier drift is exactly the kind of context that saves the next developer from reintroducing this bug

**[question] `gapFillNonces` scope in the forward-bump log** (`src/durable-objects/nonce-do.ts:~5677`)
The new log line adds `gapFillNonces: gapFillNonces.length` inside the forward-bump `if` block. This log fires *before* the gap-fill section runs (it's in the block that just replaced the early return). Does `gapFillNonces` get populated earlier in the function — e.g., from `nonceInfo.detected_missing_nonces`? If it's declared with `const` inside the gap-fill section below this block, TypeScript's TDZ would catch it at compile time. But if it's a `let` initialized to `[]` and reassigned later, this would log a stale empty array for the forward-bump case. Worth verifying the logged value is meaningful, not just zero.

**[suggestion] Guard against `last_executed_tx_nonce === -1`**
The Stacks API documentation allows `-1` for accounts with no confirmed transactions on some node versions (not just `null`). The current guard `!== null` would compute `confirmedFrontier = -1 + 1 = 0` for such a response, which happens to be the correct nonce for a fresh wallet. But it's accidental correctness — if Hiro ever returns `-1` for a different reason on an active wallet, the frontier resets to 0. Consider:
```suggestion
const confirmedFrontier = (info.last_executed_tx_nonce !== null && info.last_executed_tx_nonce >= 0)
  ? info.last_executed_tx_nonce + 1
  : info.possible_next_nonce;
```

**Operational context:**
We have direct production exposure to this bug. Wallet 0 (`SP1PMPPVCMVW96FSWFV30KJQ4MNBMZ8MRWR3JWQ7`) is currently sitting at frontier=1575 vs last_executed=1548 with 4 confirmed missing nonces [1549, 1553, 1555, 1559] — exactly the state described in the PR. The relay self-reports `healthy: true` while 18 mempool txs are queued behind unresolvable gaps. This has been escalated.

The limitation note is correct and important: after deploying this fix, the DO persistent storage still holds frontier=1575. The `POST /fill-gaps/0` admin call will be needed immediately post-deploy to clear the existing gaps. This fix prevents the condition from recurring, but does not self-heal the current state.

whoabuddy (commented):
Let's hold on merging this for now.

After further investigation, the root cause of the stuck nonces is sender-side RBF orphaning sponsor nonces — not frontier drift. When a sender replaces their own tx, the original sponsored tx gets dropped but the sponsor's nonce remains stuck in the node's mempool. `fillGapNonce` then fails with `ConflictingNonceInMempool` because it doesn't attempt RBF at a higher fee.

Two concerns with this PR in that context:

1. **The frontier change could reduce throughput.** Setting frontier to `last_executed + 1` instead of `possible_next_nonce` keeps the frontier behind mempool txs. Since headroom = `chainFrontier + CHAINING_LIMIT - assignmentHead`, a lower frontier shrinks headroom faster — the relay could start rejecting new txs prematurely even when those "gap" nonces are actually pending in the mempool and fine.

2. **It doesn't address the actual problem.** The stuck nonces need RBF recovery in `fillGapNonce`, not frontier repositioning. The early return removal (letting gap-fill run after forward-bump) is the only clearly valuable part, but it still won't help if `fillGapNonce` itself gives up on `ConflictingNonceInMempool`.

The real fix is tracked in #304 — enhance `fillGapNonce` to look up the conflicting tx's fee and retry at `fee + 1`. Once that's in place, we can revisit whether the frontier calculation matters.

PR #311, opened 2026-04-06
refactor(nonce): replace degradation state machines with per-nonce tracking and isOrigin routing

whoabuddy:
## Summary

- **Parse `is_origin` from TooMuchChaining** — stacks-core's `reason_data.is_origin` field now extracted from broadcast rejections, distinguishing origin (sender) vs sponsor chaining. Origin-side chaining no longer penalizes sponsor wallets
- **Remove 4 degradation state machines** — circuit breaker, ghost degraded, chaining degraded, and cascade detection all removed from NonceDO (~460 lines deleted)
- **Available-headroom wallet selection** — `assignNonce()` now selects wallet with highest `available = CHAINING_LIMIT - inFlightCount` instead of round-robin with degradation skips

### Evidence base
- 72h production log review showed: 34 circuit breaker triggers from only 15 actual conflicts (threshold=1 too sensitive), 0 ghost/chaining degraded triggers, 11 TooMuchChaining events all likely origin-side
- Full analysis in `.planning/2026-04-05-wallet-state-schemas/phases/03-log-review.md`

### What's removed
- `AllWalletsDegradedError`, `CIRCUIT_BREAKER_QUARANTINE_THRESHOLD`, `CIRCUIT_BREAKER_WINDOW_MS`, `GHOST_FAILURE_THRESHOLD`, `CASCADE_DETECTION_THRESHOLD`, `ALARM_INTERVAL_CASCADE_MS`, `STUCK_TX_AGE_CASCADE_MS`, `STUCK_TX_AGE_CIRCUIT_OPEN_MS`
- Methods: `recordQuarantineEvent()`, `checkCascadeThreshold()`, `isCascadeActive()`, `getEffectiveStuckTxAge()`, `incrementGhostFailures()`, `resetGhostState()`
- KV keys: `wallet_quarantine_recent:*`, `ghost_failures:*`, `ghost_degraded:*`, `cascade_quarantine_window`

### What's simplified
- `getPoolHealth()`, `buildSponsorStatusSnapshot()` — sync, pure available math
- `getObservableNonceState()` — `healthy` based on gaps + available only
- Alarm interval — active (60s) or idle (5min), no cascade mode
- Stuck tx age — constant `STUCK_TX_AGE_MS` (15min) everywhere

Supersedes #310. Implements adoption items 1-3 from the wallet-state-schemas quest.
Closes #291, closes #294, closes #302.

## Test plan

- [x] `npm run check` — typecheck passes
- [x] `npx vitest run` — 88 tests pass (11 files)
- [x] `npm run deploy:dry-run` — build succeeds
- [ ] Deploy to staging and monitor logs for 72h
- [ ] Run `npm run test:relay -- https://x402-relay.aibtc.dev` against staging

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "x402-sponsor-relay",
       "version": "1.27.2",
       "dependencies": {
-        "@aibtc/tx-schemas": "^0.3.0",
+        "@aibtc/tx-schemas": "^0.4.0",
         "@noble/curves": "^2.0.1",
         "@scure/btc-signer": "^2.0.1",
         "@stacks/common": "^7.3.1",
@@ -28,9 +28,9 @@
       }
     },
     "node_modules/@aibtc/tx-schemas": {
-      "version": "0.3.0",
-      "resolved": "https://registry.npmjs.org/@aibtc/tx-schemas/-/tx-schemas-0.3.0.tgz",
-      "integrity": "sha512-pXzW9TnmFR3uJMfajbUdAYiPKRkNlzWWL2WGZ554NAeVIPq9vWyL6x8nrYtaDohuHlZRmMj1tSVeFap9AA+adw==",
+      "version": "0.4.0",
+      "resolved": "https://registry.npmjs.org/@aibtc/tx-schemas/-/tx-schemas-0.4.0.tgz",
+      "integrity": "sha512-51z8O+mNGnqwoisKZCvvC8ZNBv0Z+RIfSTQ2CV99NFk4Z4ySp+KX+kPEpwlbLLul3AySn/AD9EHWXrjcGaAYAg==",
       "license": "MIT",
       "dependencies": {
         "zod": "^4.3.6"
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "wrangler": "^4.75.0"
   },
   "dependencies": {
-    "@aibtc/tx-schemas": "^0.3.0",
+    "@aibtc/tx-schemas": "^0.4.0",
     "@noble/curves": "^2.0.1",
     "@scure/btc-signer": "^2.0.1",
     "@stacks/common": "^7.3.1",
--- a/src/__tests__/ghost-wallet-fee-escalation.test.ts
+++ b/src/__tests__/ghost-wallet-fee-escalation.test.ts
@@ -8,169 +8,56 @@ afterEach(() => {
 // ---------------------------------------------------------------------------
 // Constants (mirrored from nonce-do.ts internals)
 // ---------------------------------------------------------------------------
-const GHOST_FAILURE_THRESHOLD = 5;
 const GAP_FILL_FEE = 30_000n;
 const MAX_BROADCAST_FEE = 90_000n;
 const CHAINING_LIMIT = 20;
 
 // ---------------------------------------------------------------------------
-// Helpers
+// buildSponsorStatusSnapshot: available math (no ghost-degraded flags)
 // ---------------------------------------------------------------------------
 
-/** Minimal nonce_state KV store for ghost counter tests */
-function makeStateStore(initial: Record<string, number> = {}) {
-  const store = new Map<string, number | null>();
-  for (const [k, v] of Object.entries(initial)) {
-    store.set(k, v);
-  }
-  return {
-    getStateValue: (key: string) => store.get(key) ?? null,
-    setStateValue: (key: string, value: number) => store.set(key, value),
-    walletGhostFailuresKey: (NonceDO as any).prototype.walletGhostFailuresKey,
-    walletGhostDegradedKey: (NonceDO as any).prototype.walletGhostDegradedKey,
-    _store: store,
-  };
-}
-
-// ---------------------------------------------------------------------------
-// Ghost counter semantics
-// ---------------------------------------------------------------------------
-
-describe("ghost wallet counter semantics", () => {
-  const incrementGhostFailures = (NonceDO as any).prototype.incrementGhostFailures;
-  const resetGhostState = (NonceDO as any).prototype.resetGhostState;
-  const walletGhostFailuresKey = (NonceDO as any).prototype.walletGhostFailuresKey;
-  const walletGhostDegradedKey = (NonceDO as any).prototype.walletGhostDegradedKey;
-
-  it("increments ghost failures and marks degraded at threshold", () => {
-    const ctx = { ...makeStateStore(), log: vi.fn() };
-
-    // Increment up to threshold - 1: NOT degraded yet
-    for (let i = 1; i < GHOST_FAILURE_THRESHOLD; i++) {
-      const becameDegraded = incrementGhostFailures.call(ctx, 0);
-      expect(becameDegraded).toBe(false);
-      expect(ctx.getStateValue(walletGhostFailuresKey.call(ctx, 0))).toBe(i);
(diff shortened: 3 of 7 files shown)


arc0btc (approved):
Removes 4 overlapping nonce control state machines backed by 72h production log evidence, and adds correct `is_origin` routing so origin-side TooMuchChaining no longer penalizes sponsor wallets. The evidence base makes this deletion easy to trust.

**What works well:**
- The production data justification is rigorous: 34 circuit breaker triggers from 15 actual conflicts, 0 ghost/chaining degraded triggers, 11 TooMuchChaining events all origin-side. Deletion with evidence is the right call.
- `is_origin` parsing in `settlement.ts` is clean — parses `reason_data.is_origin` from node response, propagates via `isOriginChaining` through `BroadcastAndConfirmResult` and `BroadcastOnlyResult`, correctly sets `retryable: !isOrigin` and uses a distinct `terminalReason: "origin_chaining_limit"`.
- `getPoolHealth()` and `buildSponsorStatusSnapshot()` are now sync — removing N async KV reads (quarantine list per wallet) from what were likely hot paths.
- `assignNonce()` loses the per-wallet `walletQuarantineRecentKey` storage read on the critical path. Real win.
- The test cleanup mirrors the code cleanup: ghost counter semantics tests → available math tests. Accurate and sufficient.

**[suggestion] Origin-side TooMuchChaining still retries in `queue-consumer.ts`** (`src/queue-consumer.ts:300`)

The retry gate checks `isTooMuchChaining` without excluding `isOriginChaining`:
```typescript
if ((isNonceConflict || isTooMuchChaining) && attempt < MAX_ATTEMPTS) {
```
But `settlement.ts` correctly returns `retryable: !isOrigin`. Retrying origin-side chaining doesn't help — the agent's address is congested and won't clear until their pending transactions confirm. The retries consume queue slots without any chance of success until MAX_ATTEMPTS is exhausted.

```suggestion
if ((isNonceConflict || (isTooMuchChaining && !isOriginChaining)) && attempt < MAX_ATTEMPTS) {
```

This fails fast for origin-side chaining while keeping retry behavior for sponsor-side congestion and nonce conflicts.

**[suggestion] Stale KV keys need a cleanup path**

Old state machines wrote to `wallet_ghost_failures:*`, `wallet_ghost_degraded:*`, `wallet_quarantine_recent:*`, and `cascade_quarantine_window` in KV. These keys are now orphaned — no code reads them, but they accumulate in storage. Worth either:
- A one-shot cleanup in the alarm handler (first invocation after deploy clears the known-prefix keys)
- Or at minimum, a comment in the code or a migration note in the PR

Not a correctness issue, but ops debt that'll show up in a storage audit.

**[nit] Hardcoded booleans in `getPoolHealth()` return**

```typescript
return { allWalletsDegraded: false, totalReserved, totalCapacity, wallets };
```

The `wallets` array also returns `circuitBreakerOpen: false` and `quarantineCount: 0` hardcoded. This is correct (no wallets are degraded anymore) but could confuse consumers who are monitoring these fields. A brief comment like `// circuit breaker removed; always false for API compatibility` would clarify intent and prevent future "why is this always false?" confusion.

**Code quality notes:**
- The `degradedWallets` tracking and `AllWalletsDegradedError` removal are clean — no orphaned catch clauses left.
- Comments stripped from `releaseNonce()` were doing real work explaining the branching logic. The branch structure is still readable without them, but the "why a gap gets created" context is now only findable in git history.
- Minor: one extra blank line left in `reconcileNonceForWallet` after the `resetGhostState` removal.

**Operational context:** We've been running against this relay 24/7 and have lived through the circuit breaker over-triggering in production. The 34-from-15 ratio matches what we observed in our own nonce incident logs. Removing threshold=1 is clearly the right call. The `is_origin` distinction matters for our use case directly — after outages, Arc's own address sometimes has stale pending transactions, which previously cascaded into sponsor wallet degradation. This PR fixes the root misattribution.

One question: the `@aibtc/tx-schemas` bump from `^0.3.0` → `^0.4.0` — what schema additions does 0.4.0 include beyond the `isOriginChaining` field? Asking in case there are other new fields we should handle on the relay side.

arc0btc (approved):
Replaces four degradation state machines with simpler per-nonce tracking and adds `isOrigin` routing for TooMuchChaining — a net deletion of ~460 lines backed by 72h production log analysis.

**What works well:**
- The evidence base is rigorous: 34 circuit-breaker triggers from 15 real conflicts (threshold=1 too sensitive) + 0 ghost/chaining-degraded triggers in 72h = strong signal that the machinery had zero actual benefit. Removing it based on data is the right call.
- `isOrigin` extraction from `reason_data.is_origin` is clean and the strict `=== true` check is safe against truthy coercion from external API responses.
- `getPoolHealth()` and `buildSponsorStatusSnapshot()` converted from async KV reads to sync — these are called on every status request, so this is a meaningful performance win.
- `cleanupLegacyDegradationKeys()` is a well-structured one-shot migration: flag-gated, logs what it deleted, handles both KV (`wallet_quarantine_recent:*`, `cascade_quarantine_window`) and nonce_state SQL (`wallet_ghost_*` zeroed via `setStateValue`). Good hygiene.
- `circuitBreakerOpen: false` hardcoded with `@deprecated` JSDoc preserves API compatibility without polluting the logic — the right tradeoff for a live API.
- Test file trimmed from 169 to 56 lines and rewritten around the actual behavior (available math) rather than the removed state machines — tests are now testing real invariants.

**[question] `recommendation` fallback behavior when capacity is exhausted without gaps** (`src/durable-objects/nonce-do.ts`)

Old: `!healthy && (anyGaps || allDegraded)` — fallback recommended when circuit-breakers were open  
New: `!healthy && anyGaps` — fallback only recommended when gaps exist

This means: if all wallets are at chaining capacity (`totalAvailable === 0`, no gaps), `healthy: false` but `recommendation: null`. Callers that relied on the pre-flight `/nonce-state` recommendation to route around a fully-saturated relay will now get `recommendation: null` and only learn of the condition via the 503 `ChainingLimitError`. Is that intentional? Happy to be corrected if callers only act on the 503 path — just wanted to surface it.

**[suggestion] Ghost state key cleanup completeness** (`cleanupLegacyDegradationKeys()`)

The KV keys (`wallet_quarantine_recent:*`, `cascade_quarantine_window`) are deleted via `state.storage.delete()`. The nonce_state SQL keys (`wallet_ghost_failures:*`, `wallet_ghost_degraded:*`) are zeroed via `setStateValue` — they remain as zero rows rather than being absent. Not a bug since all read paths check `=== 1` (now removed), but if the nonce_state table grows large it might be worth using `deleteStateValue` to clean them entirely rather than leave zero-value rows. Low priority.

**Code quality notes:**
- `ObservableWalletState` interface still carries `circuitBreakerOpen` (deprecated) but `ghostDegraded` and `ghostFailures` fields are fully removed from the return value — this is a breaking change to the `/nonce-state` endpoint. If any consumers read those fields they'll get `undefined`. Presumably intentional given #302 closes here, but worth noting for changelog.
- `assignNonce()` wallet selection loop simplified from a while+attempts counter to a clean for loop — much easier to audit. Good call.

**Operational context:**
We run against this relay in production (~80 x402 messages/day). Our monitoring ([WATCH] in MEMORY) currently shows nonce gaps [2920, 2921] on the sponsor wallet — the per-nonce occupied tracking that this PR leaves intact is what handles gap resolution. The removal of cascade-aware gap-fill limits (`effectiveMaxGapFills = cascadeActive ? 15 : 5` → constant 5) is fine for our load; cascade mode never triggered anyway per the log analysis.

The simplification is justified and the implementation is clean. Minor questions above — neither is blocking.

PR #300, opened 2026-04-04
fix(security): upgrade lodash to 4.18.0 (CVE-2026-4800)

arc0btc:
## Summary

- Patches **CVE-2026-4800** / GHSA-r5fr-rjxr-66jc (CVSS 8.1, high severity)
- `lodash < 4.18.0` allows code injection via `_.template` when untrusted input is passed as `options.imports` key names
- lodash is a **transitive** dependency (pulled in by `async`); this repo does not call `_.template` directly — exploitability is low, but patching eliminates the risk entirely

## Changes

- `package-lock.json`: bump lodash 4.17.23 → 4.18.0 with correct integrity hash
- `package.json`: add `overrides.lodash: 4.18.0` to pin the transitive version on future installs

## Test plan

- [ ] CI type-check passes (`tsc --noEmit`)
- [ ] Dependabot alert #22 resolves after merge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #315, opened 2026-04-06
fix(deps): pin lodash transitive dependency to 4.18.1

whoabuddy:
## Summary

- Adds `overrides.lodash` in `package.json` to pin the transitive dep at 4.18.1
- Ensures future `npm install` won't regress below the CVE-2026-4800 fix
- Extracted from #300 (updated to 4.18.1 to match Dependabot's #299)

## Test plan

- [ ] `npm install` resolves lodash to 4.18.1
- [ ] `npm run check` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,9 @@
     "vitest": "^4.0.18",
     "wrangler": "^4.75.0"
   },
+  "overrides": {
+    "lodash": "4.18.1"
+  },
   "dependencies": {
     "@aibtc/tx-schemas": "^0.4.0",
     "@noble/curves": "^2.0.1",


arc0btc (approved):
Pins lodash transitive dependency to 4.18.1 to address CVE-2026-4800 — correct mechanism for forcing transitive dep versions in npm v7+.

**What works well:**
- `overrides.lodash` is the right npm approach here (vs `resolutions` which is yarn/pnpm-specific)
- Exact pin `"4.18.1"` gives deterministic resolution — appropriate for a security fix where regression prevention is the goal
- Minimal, isolated change extracted from #300 to keep the security fix reviewable on its own
- Cloudflare Workers deployed successfully to both staging and production (confirmed in CI comments above), so the change is already validated in the build pipeline

**[question] Exact vs range pin** (`package.json`)
The exact `"4.18.1"` pin means future lodash patches won't auto-apply. This is intentional for a security pin, but it means future lodash releases will require manual `overrides` updates (or Dependabot bumping this entry). Not blocking — just worth tracking so the override doesn't become stale against a future CVE.

**Code quality notes:**
- [nit] `overrides` is placed between `devDependencies` and `dependencies`. No functional impact, but some teams prefer a consistent placement (before `dependencies` or at the end). Cosmetic only.

**Operational context:**
We run this relay in production for Arc's x402 welcome flow. Relay v1.27.2 is currently operational and healthy (throughput stable, nonce gap cleared). Lodash CVEs typically target prototype pollution — fixing the transitive dep promptly in a Cloudflare DO execution context is exactly right. This unblocks our production security posture without any behavioral changes.

tfireubs-ui (approved):
LGTM. Minimal 3-line npm overrides pin for lodash 4.18.1 — prevents CVE-2026-4800 regression on future installs. Clean and correct.

PR #316, opened 2026-04-06
fix(health): derive status from nonce pool health

whoabuddy:
## Summary

- `/health` status now reflects actual nonce pool health instead of hardcoded `"ok"`
- Returns `"degraded"` when the pool reports unhealthy (circuit breaker, gaps, low capacity)
- Gracefully stays `"ok"` when the coordinator is unreachable
- Adds tests for status derivation logic and schema enum documentation

Extracted from #300 (health endpoint changes only, separated from the lodash fix).

## Test plan

- [x] `npm run check` passes
- [x] All 92 tests pass (5 health tests including 3 new derivation tests)

### DIFF
--- a/src/__tests__/health.test.ts
+++ b/src/__tests__/health.test.ts
@@ -12,4 +12,36 @@ describe("Health schema", () => {
     expect(properties).toHaveProperty("version");
     expect(properties).not.toHaveProperty("nonce");
   });
+
+  it("status field documents ok and degraded enum values", () => {
+    const endpoint = new Health();
+    const statusProp =
+      endpoint.schema.responses["200"].content["application/json"].schema.properties.status;
+    expect(statusProp.enum).toEqual(["ok", "degraded"]);
+  });
+});
+
+/**
+ * Tests for the status derivation logic used in Health.handle().
+ *
+ * The handler derives status from poolHealthy (boolean | null):
+ * - poolHealthy === false → "degraded"
+ * - poolHealthy === true or null (unavailable) → "ok"
+ */
+describe("Health status derivation", () => {
+  function deriveStatus(poolHealthy: boolean | null): "ok" | "degraded" {
+    return poolHealthy === false ? "degraded" : "ok";
+  }
+
+  it("returns 'ok' when pool is healthy", () => {
+    expect(deriveStatus(true)).toBe("ok");
+  });
+
+  it("returns 'degraded' when pool is unhealthy", () => {
+    expect(deriveStatus(false)).toBe("degraded");
+  });
+
+  it("returns 'ok' when pool state is unavailable (null)", () => {
+    expect(deriveStatus(null)).toBe("ok");
+  });
 });
--- a/src/endpoints/health.ts
+++ b/src/endpoints/health.ts
@@ -5,6 +5,10 @@ import { VERSION } from "../version";
 /**
  * Health check endpoint
  * GET /health
+ *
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Derives `/health` status from actual nonce pool health rather than hardcoding `"ok"` — directly addresses the gap where our health monitoring gave false confidence about relay state.

**What works well:**
- Graceful fallback: coordinator unreachable → `"ok"` is correct behavior for a health endpoint (never fail, just degrade signal quality). The try/catch is the right shape here.
- `poolHealthy === false` (not `=== null`) mapping to `"degraded"` is precise — null (unavailable) stays `"ok"`, which is intentional.
- Schema enum with the consumer guidance ("treat any non-'ok' value as unhealthy") is good API contract documentation.
- Clean extraction from #300. Focused scope makes this easy to review and merge independently.

**[suggestion] Cache the DO fetch for high-frequency callers** (`src/endpoints/health.ts`)
Every `/health` call now does a round-trip to the nonce DO. Load balancers and uptime monitors hit `/health` at high frequency — Cloudflare Workers internal latency is low but it adds up. A 15–30 second in-memory cache for `poolHealthy` would reduce DO calls by 99%+ under typical monitoring load:
```suggestion
  async handle(c: AppContext) {
    let poolHealthy: boolean | null = null;

    if (c.env.NONCE_DO) {
      try {
        const stub = c.env.NONCE_DO.get(c.env.NONCE_DO.idFromName("sponsor"));
        const response = await stub.fetch("https://nonce-do/nonce-state");
        if (response.ok) {
          const state = (await response.json()) as { healthy?: boolean };
          poolHealthy = state.healthy ?? null;
        }
      } catch {
        // Coordinator unreachable — stay "ok", logged separately by /nonce/state
      }
    }
```
(Not blocking — the cost is negligible at current throughput and the pattern is simpler without caching.)

**[question] Only "sponsor" nonce instance?** (`src/endpoints/health.ts:55`)
The code fetches `NONCE_DO.idFromName("sponsor")`. Is "sponsor" always the canonical instance name? We've seen both sender and sponsor nonce contexts in operational logs. If there are multiple DO instances, should health check any of them or specifically just "sponsor"? Fine if "sponsor" is the only relevant one here, just confirming.

**[nit] Test helper duplication** (`src/__tests__/health.test.ts`)
The `deriveStatus` helper in the test file mirrors the expression in `health.ts`. If the logic ever changes, both places need updating. Not a concern at this complexity level, but worth noting if the derivation grows.

**Code quality notes:**
- Logic is minimal and clear — no abstraction needed.
- The inline comment `// Coordinator unreachable — logged separately by /nonce/state` is helpful for future maintainers wondering why errors are silently swallowed.

**Operational context:**
We run `arc skills run --name bitcoin-wallet -- check-relay-health` against this relay and monitor nonce pool state in our sensors. With this change, our health check callers get accurate degradation signal instead of false `"ok"` — directly improves our stale-lock detection and relay monitoring. The nonce gap patterns we've seen ([1621] recently resolved, [1739] emerging) are exactly the cases where `"degraded"` status would have been valuable to surface. Solid improvement.

tfireubs-ui (approved):
LGTM. Correctly derives /health status from nonce pool health — returns 'degraded' when pool reports unhealthy, defaults to 'ok' when coordinator unreachable. Tests cover all three cases (healthy, unhealthy, unavailable). Clean extraction from #300.

PR #326, opened 2026-04-08
fix(nonce): move Hiro fetches outside blockConcurrencyWhile to prevent cold start crashes (closes #324)

tfireubs-ui:
## Problem

`NonceDO` crashes intermittently on cold start (and during alarm reconciliation) because `assignNonce()` and `alarm()` both hold a `blockConcurrencyWhile` lock while making sequential HTTP calls to the Hiro API. With `SPONSOR_WALLET_COUNT=10`, up to 10 sequential Hiro fetches can occur inside a single lock window — easily exceeding Cloudflare's ~30 s `blockConcurrencyWhile` limit and crashing the DO, dropping all queued requests.

### `reconcileNonceForWallet()`
Added an optional `prefetchedNonceInfo?: HiroNonceInfo | null` parameter:
- `HiroNonceInfo` → use the pre-fetched result, skip HTTP call
- `null` → Hiro was unreachable during pre-fetch, return null silently (existing "skip" behavior)
- `undefined` → no pre-fetch (backward-compatible; falls back to live fetch)

### `alarm()`
Split into two phases:
1. **Outside lock (Phase 1):** fetch the list of initialized wallets, compute the reconcile slice, then pre-fetch all wallet nonce infos in **parallel** via `Promise.all` before entering `blockConcurrencyWhile`.
2. **Inside lock (Phase 2):** all state mutations. Pass pre-fetched `HiroNonceInfo` into `reconcileNonceForWallet` — no Hiro I/O occurs inside the lock.

### `assignNonce()`
Split into two phases:
1. **Outside lock (Phase 1):** identify wallets that need initialization (no head seeded) or whose lookahead-guard cache has expired. Pre-fetch their nonce infos in **parallel** via `Promise.all`.
2. **Inside lock (Phase 2):** apply the pre-fetched seeds directly (no HTTP). Also updates the `hiroNonceCache` from pre-fetched data so the subsequent `fetchNextNonceForWallet` call is served from cache (no additional Hiro call). Falls back to a live fetch only when the pre-fetch genuinely failed (e.g., Hiro down at cold-start on a brand-new DO).

## Result

- `blockConcurrencyWhile` callbacks in both `alarm()` and `assignNonce()` no longer make any Hiro HTTP calls in the normal (post-first-tick) path.
- Cold-start with N=10 wallets: 10 parallel Hiro fetches complete well within 30 s, then the lock is taken for a brief SQL-only mutation window.
- Hiro unreachability is handled fail-open: wallets that couldn't be pre-fetched either skip reconciliation (alarm) or fall back to a single live fetch (assignNonce) as before.

## Testing

TypeScript check passes with no new errors (`tsc --noEmit --skipLibCheck`). The pre-existing `worker-configuration.d.ts` error is unrelated to this change and exists on `main`.

closes #324

### DIFF
--- a/src/durable-objects/nonce-do.ts
+++ b/src/durable-objects/nonce-do.ts
@@ -3785,6 +3785,54 @@ export class NonceDO {
       throw new Error("Missing sponsor address");
     }
 
+    // -------------------------------------------------------------------------
+    // Phase 1 (outside lock): Pre-fetch Hiro nonce data for any wallets that
+    // need initialization or whose lookahead-guard cache has expired. Hiro I/O
+    // must NOT happen inside blockConcurrencyWhile — with N wallets, sequential
+    // fetches inside the lock can exceed Cloudflare's ~30 s limit.
+    // -------------------------------------------------------------------------
+
+    // Compute effectiveWalletCount from in-memory SQL state (no I/O).
+    const storedDynamicPre = this.getStateValue("dynamic_wallet_count");
+    const effectiveWalletCountPre = Math.max(
+      1,
+      Math.min(
+        Math.max(walletCount, storedDynamicPre ?? 0),
+        this.getSponsorWalletMax()
+      )
+    );
+    const resolveAddressPre = (wi: number): string =>
+      addresses?.[String(wi)] ?? sponsorAddress;
+
+    // Collect wallet indices that need a fresh Hiro fetch:
+    //   - unseeded (ledgerGetWalletHead === null): must fetch to initialize
+    //   - lookahead cache expired or missing: needed for stale-head / cap guards
+    const walletsFetchNeeded: number[] = [];
+    for (let i = 0; i < effectiveWalletCountPre; i++) {
+      const needsSeed = this.ledgerGetWalletHead(i) === null;
+      const cached = this.hiroNonceCache.get(i);
+      const cacheExpired = !cached || Date.now() >= cached.expiresAt;
+      if (needsSeed || cacheExpired) {
+        walletsFetchNeeded.push(i);
+      }
+    }
+
+    // Pre-fetch in parallel — null entry means Hiro was unreachable for that wallet.
+    const prefetchMap = new Map<number, HiroNonceInfo | null>();
+    await Promise.all(
+      walletsFetchNeeded.map(async (wi) => {
+        try {
+          const info = await this.fetchNonceInfo(resolveAddressPre(wi));
+          prefetchMap.set(wi, info);
+        } catch (_e) {
+          prefetchMap.set(wi, null);
+        }
+      })
+    );
+
+    // -------------------------------------------------------------------------
+    // Phase 2 (inside lock): All state reads and mutations. No Hiro I/O.
+    // -------------------------------------------------------------------------
     return this.state.blockConcurrencyWhile(async () => {
       const currentAlarm = await this.state.storage.getAlarm();
       if (currentAlarm === null) {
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Moves Hiro I/O out of `blockConcurrencyWhile` in both `assignNonce()` and `alarm()` — the right fix for this class of DO crash.

**What works well:**
- Two-phase split is clean and well-documented. Phase 1 parallelizes Hiro fetches; Phase 2 is pure state mutation. The intent is immediately clear from the section comments.
- `reconcileNonceForWallet(prefetchedNonceInfo?: HiroNonceInfo | null)` — the three-state parameter (`undefined` = no pre-fetch, `null` = pre-fetch failed, `HiroNonceInfo` = success) is a bit unusual but the fallback semantics are correct and documented.
- Fail-open behavior preserved: if Hiro is unreachable during pre-fetch, wallets either skip reconciliation (`alarm`) or fall back to a live fetch (`assignNonce`). No regression in Hiro-down handling.
- Cache warming as a side effect of pre-fetch (updating `hiroNonceCache` for already-seeded wallets) is a nice bonus — means `fetchNextNonceForWallet` inside the lock hits cache, not Hiro.

**[question] Reconcile slice mismatch in `alarm()`** (`src/durable-objects/nonce-do.ts`)
Phase 1 reads `walletCursorPre` and computes `reconcileWalletsPre` to determine which wallets to pre-fetch. Phase 2 re-reads `walletCursor` inside the lock and computes `reconcileWallets` independently. If another operation advances the cursor between Phase 1 and Phase 2, the two slices diverge — wallets in Phase 2 but not Phase 1 won't have entries in `prefetchedNonceInfos`, and `reconcileNonceForWallet` falls back to a live Hiro fetch inside the lock.

In a DO with `blockConcurrencyWhile`, this race is essentially impossible (single-threaded model, and `alarm()` owns the cursor advance). Just worth confirming that `getStateValue(ALARM_WALLET_CURSOR_KEY)` reads from in-memory state and not Cloudflare storage (which could theoretically return a stale value). Is this intentional / do you have confidence the slices stay in sync?

**[nit] Redundant has+get pattern** (`src/durable-objects/nonce-do.ts`)
```suggestion
const prefetched = prefetchedNonceInfos.get(walletIndex);
```
`Map.get` returns `undefined` when the key is absent and `null` when present with a null value — same semantics as the `has + get` check, one fewer lookup.

**Operational context:** We run this relay in production (v1.27.3) and have been tracking nonce gaps intermittently. Cold-start crashes causing dropped queued requests would explain some of the gap patterns we've seen — specifically the multi-wallet nonce conflicts that appear after relay restarts. This fix directly addresses the root cause. The parallel pre-fetch design matches how our own sensor concurrency works. Approving.

whoabuddy:
**Addressing arc0btc's review questions:**

**[question] Cursor slice mismatch in `alarm()`** — confirmed safe. `getStateValue(ALARM_WALLET_CURSOR_KEY)` reads from in-memory SQL state (`this.sql`), not Cloudflare durable storage — the comment on line 6729 ("read from SQL (fast, no I/O)") documents this. Cloudflare DOs are single-threaded: between the Phase 1 cursor read and the Phase 2 cursor re-read, no other operation can run, so `walletCursor == walletCursorPre` is guaranteed. The cursor is only advanced at the *end* of Phase 2, well after the reconcile loop. The re-computation in Phase 2 is intentionally defensive — if someone later extracts the reconcile block into its own helper, it remains self-contained. The slices are always identical.

**[nit] Redundant `has + get` pattern** — applied in commit `e918e94`. Replaced:
```ts
// before
const prefetched: HiroNonceInfo | null | undefined = prefetchedNonceInfos.has(walletIndex)
  ? prefetchedNonceInfos.get(walletIndex)
  : undefined;
```
with:
```ts
// after
const prefetched = prefetchedNonceInfos.get(walletIndex);
```
`Map.get` already returns `undefined` for absent keys and `null` for keys set to `null` — same semantics, one fewer lookup. Also updated the inline comment to document the `undefined`/`null` semantics explicitly.

Local CI: `npm run check` and `npm run deploy:dry-run` both pass with no new errors after the nit fix.

PR #332, opened 2026-04-10
fix(deps): pin axios>=1.15.0 to patch CVE-2025-62718 (SSRF)

arc0btc:
## Summary

- Pins `axios` to `>=1.15.0` via npm `overrides` to patch CVE-2025-62718
- axios was a transitive dependency via `x402-stacks` at version 1.13.6 (vulnerable)
- After override: package-lock.json resolves axios to 1.15.0

## Vulnerability

**CVE-2025-62718** — axios NO_PROXY Hostname Normalization Bypass (SSRF)

Axios did not correctly handle hostname normalization when checking `NO_PROXY` rules. Requests to loopback addresses like `localhost.` (trailing dot) or `[::1]` (IPv6 literal) could skip `NO_PROXY` matching and route through a configured proxy.

**Impact on this service**: Low — this runs as a Cloudflare Worker where NO_PROXY settings don't apply. However, patching is correct practice and resolves the Dependabot critical alert.

## Test plan

- [x] `npm install --package-lock-only` regenerated lock file cleanly
- [x] `package-lock.json` confirms axios resolves to `1.15.0`
- [ ] CI type-check passes

Closes #31 (Dependabot alert)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1961,14 +1961,14 @@
       "license": "MIT"
     },
     "node_modules/axios": {
-      "version": "1.13.6",
-      "resolved": "https://registry.npmjs.org/axios/-/axios-1.13.6.tgz",
-      "integrity": "sha512-ChTCHMouEe2kn713WHbQGcuYrr6fXTBiu460OTwWrWob16g1bXn4vtz07Ope7ewMozJAnEquLk5lWQWtBig9DQ==",
+      "version": "1.15.0",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-1.15.0.tgz",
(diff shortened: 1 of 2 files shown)


tfireubs-ui (approved):
LGTM — same axios 1.13.6 → 1.15.0 CVE-2025-62718 fix as mcp-server #459. Clean dep bump, no code changes. Also picks up proxy-from-env 2.1.0.

whoabuddy:
Superseded by #333. Axios only entered the lockfile via the unused x402-stacks dependency, so the tighter fix is to remove that package instead of keeping an override for code we do not use.

PR #363, opened 2026-04-30
fix(stream): improve Hiro tx stream error logging and reduce stream budget (closes #313)

whoabuddy:
## Summary

Closes #313. Two related fixes in the Hiro WebSocket tx-stream path.

### 1. Error logging — `{}` → real diagnostics

`hiro-tx-stream.ts:230` was logging `JSON.stringify(event)` on every WebSocket `error` event, which always produced `"{}"` because the Cloudflare Workers WebSocket error event has non-enumerable properties. **Zero diagnostic visibility on stream failures.** (Issue body: 13 occurrences in a single day with empty `{}` payloads.)

Extract the `type` / `code` / `message` fields the runtime exposes, mirroring the existing `handleClose` handler. When no fields are available, log a clear sentinel string instead of `{}`.

```ts
const err = event as { message?: string; type?: string; code?: number };
logger.warn("Hiro tx stream errored; falling back", {
  txid,
  type: err?.type ?? null,
  code: err?.code ?? null,
  message: err?.message ?? "(WebSocket error event with no enumerable properties)",
});
```

### 2. Budget allocation — swap stream vs REST split

`settlement.ts:650-655` allocated **150s to the stream and 30s to REST polling** out of a 180s effective budget. In practice the Hiro WebSocket either returns terminal status within seconds or errors fast. Waiting 150s on a stream that errored at t=1s is wasted budget that should be REST polling.

Swap: stream gets **≤30s as a head-start**, REST polling gets the remainder (~150s in default case).

```ts
// Before: stream 150s + fallback 30s
const fallbackBudgetMs = Math.min(STREAM_FALLBACK_TAIL_MS, effective, max(10_000, effective/3));
const streamBudgetMs   = Math.max(0, effective - fallbackBudgetMs);

// After: stream ≤30s + fallback ~150s
const streamBudgetMs   = Math.min(STREAM_BUDGET_MS, effective);
const fallbackBudgetMs = Math.max(0, effective - streamBudgetMs);
```

Renamed the const `STREAM_FALLBACK_TAIL_MS` → `STREAM_BUDGET_MS` to reflect the new semantics — it's now the cap on stream wait, not the reserve for fallback.

## Test plan

- [x] `npm run check` (TypeScript) clean
- [ ] CI green
- Operationally: after merge, watch the `Hiro tx stream errored; falling back` warn-level entries to confirm `type`/`code`/`message` are now populated for typical failures (or that the sentinel fires when the runtime really gives nothing).

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/services/hiro-tx-stream.ts
+++ b/src/services/hiro-tx-stream.ts
@@ -228,9 +228,17 @@ export async function waitForHiroTxConfirmationViaStream(
     };
 
     const handleError = (event: unknown) => {
+      // Cloudflare Workers WebSocket error events have non-enumerable
+      // properties, so JSON.stringify(event) was always producing "{}".
+      // Extract the diagnostic fields the runtime exposes (mirroring the
+      // close-event handler below).
+      const err = event as { message?: string; type?: string; code?: number };
       logger.warn("Hiro tx stream errored; falling back", {
         txid,
-        error: JSON.stringify(event),
+        type: err?.type ?? null,
+        code: err?.code ?? null,
+        message:
+          err?.message ?? "(WebSocket error event with no enumerable properties)",
       });
       finish(null, "error");
     };
--- a/src/services/settlement.ts
+++ b/src/services/settlement.ts
@@ -52,8 +52,10 @@ const MAX_POLL_DELAY_MS = 8_000;
  *  After broadcasting, if the first polling round times out, we retry
  *  polling (without re-broadcasting) since the tx is already in mempool. */
 const POLL_RETRY_ROUNDS = 2;
-/** Reserve a tail window for degraded REST polling if tx streaming is unavailable. */
-const STREAM_FALLBACK_TAIL_MS = 30_000;
+/** Cap on tx-stream wait. The Hiro WebSocket either delivers a terminal
+ *  update within seconds or errors fast — there's no value in waiting longer.
+ *  REST polling is the long-tail strategy; this is just a head-start. */
+const STREAM_BUDGET_MS = 30_000;
 
 // Hiro API timeout configuration
 /** Timeout for each broadcast attempt POST to Hiro /v2/transactions (ms).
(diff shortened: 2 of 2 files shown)


arc0btc (approved):
Two targeted fixes in the Hiro WebSocket stream path — both correct and well-reasoned.

**What works well:**
- The error-event extraction in `hiro-tx-stream.ts` mirrors the existing `handleClose` handler — consistent pattern across both event types.
- The sentinel string `"(WebSocket error event with no enumerable properties)"` is a good fallback: it's unambiguous in logs, unlike the silent `{}` that was going in before.
- The budget swap in `settlement.ts` is the right call. Letting a stream that errors at t=1s sit idle for 149s before REST polling kicks in was genuinely wasteful. ≤30s head-start is appropriate given the Hiro WebSocket's behavior.
- Renaming `STREAM_FALLBACK_TAIL_MS` → `STREAM_BUDGET_MS` makes the semantics clear — the old name implied it was a reserve for the fallback, which was backwards from what the variable actually did.

**[question] Edge case at low total budgets** (`src/services/settlement.ts`)
When `effectivePollTimeMs ≤ STREAM_BUDGET_MS` (e.g., 20s total), `fallbackBudgetMs` becomes 0 — REST polling gets no time at all. The old code had the same gap, so this isn't a regression, but worth documenting. Is there a scenario where a caller passes a short `maxPollTimeMs` and expects REST fallback to be available?

**Code quality notes:**
- `err?.type` uses optional chaining on `err`, but `err` is always assigned (it's cast from `event` which is typed as `unknown` but not actually optional). The `?.` is harmless but slightly misleading — `err.type` with the interface guard would be sufficient. Not worth changing; just noting it.
- The comment block in `hiro-tx-stream.ts` is three lines where one would do: the key fact is "Cloudflare Workers WebSocket error events have non-enumerable properties." The rest is implementation detail visible from the code itself. Minor nit.

**Operational note:** We've been watching this path (`x402-relay [WATCH, v1.31.0]` — nonce gaps on the sponsor). Better error logging on stream failures will help diagnose whether the nonce gaps and stream errors are related. Appreciate the reduced fallback window too — our settlement path should be more responsive to stream failures post-merge.

Issue #373, opened 2026-05-15
/settle should re-sponsor pre-sponsored txs on ConflictingNonceInMempool instead of returning error


whoabuddy:
## Summary

`/settle` returns a hard `CONFLICTING_NONCE` error to callers whenever a pre-sponsored tx hits `ConflictingNonceInMempool` on broadcast, even though the relay has everything it needs to recover by re-sponsoring with a fresh sponsor nonce. The existing inline-resync + retry path at `src/endpoints/settle.ts:556` is gated on `sponsorNonce !== null`, which is false for the pre-sponsored-tx case, so we fall through to the warn at `src/endpoints/settle.ts:656` and emit a client error.

This is the agent-visible failure that is currently showing up as repeated `Payment settlement failed | errorReason: conflicting_nonce` in landing-page logs (18 such errors today, 2026-05-15, from a single burst at 13:02–13:03 UTC).

## Current behavior

`src/endpoints/settle.ts`:

```ts
// line ~556
if (sponsorNonce !== null && (broadcastResult.nonceConflict || broadcastResult.tooMuchChaining)) {
  // inline resync + re-sponsor + retry
  ...
}

// line ~655
if (clientRejection) {
  logger.warn(\"Broadcast rejected by node (client error)\", { ... });
  return v2Error(mapClientRejectionToV2Code(clientRejection), 200);
}
```

When `/settle` is called on a tx that was sponsored earlier (e.g., via `/sponsor` and held in a downstream queue for retry), `sponsorNonce` at `/settle` time is `null` because this endpoint did not assign the nonce. So the recovery branch is skipped and the agent receives a terminal `CONFLICTING_NONCE` reply.

## Why this matters

The inner client-signed payload is still valid — only the sponsor side is stale. We already have the logic to re-sponsor (`retrySponsorService.sponsorTransaction(validatedTxForRetry.transaction)`); we just don't reach it on the pre-sponsored path.

## Proposed change

In `src/endpoints/settle.ts`, when broadcast returns `nonceConflict: true` AND `sponsorNonce === null`:

1. Extract the inner client-signed payload from the submitted tx.
2. Call `SponsorService.sponsorTransaction(innerTx)` to mint a fresh sponsor nonce + signature.
3. Re-broadcast.
4. On success → `handleBroadcastSuccess`. On failure → schedule delayed resync + return v2Error as today.

## Evidence

[...]

PR #379, opened 2026-05-16
feat(relay): add nonceExpiresAt to /relay and /sponsor responses (#374)

### LINKED ISSUE #374: Add explicit sponsor-nonce TTL (validUntil) on /sponsor responses so downstream queues stop retrying stale hex
## Summary

The relay reserves a sponsor nonce when `/sponsor` runs and treats that reservation as expired after `STALE_THRESHOLD_MS = 10 min` (`src/durable-objects/nonce-do.ts:425`, used at `src/durable-objects/nonce-do.ts:6367-6371`). Downstream consumers (most prominently the landing-page inbox reconciliation queue) keep the sponsored tx hex alive for ~70 min of retries and have no way to know the relay has already reclaimed the slot. The result is the failure mode in #373 — pre-sponsored hex broadcast after expiry → `ConflictingNonceInMempool`.

This issue asks the relay to publish its TTL contract explicitly instead of leaving callers to guess.

## Current behavior

`/sponsor` responds with the sponsored tx hex + sponsor nonce. No expiry information. The 10-min TTL lives only in relay-internal code; consumers infer nothing about it.

Reconcile loop reclaims the slot:

```ts
// src/durable-objects/nonce-do.ts:6367
if (ageMs > STALE_THRESHOLD_MS) {
  verdict = \"expired\";
  reason = \"stale_assigned_never_broadcast\";
  ...
  this.ledgerMarkExpiredByReconcile(walletIndex, nonce, reason);
}
```

After this, the next /sponsor call can be handed the same nonce, which is what produces the conflict observed in production.

## Proposed change

1. Add `sponsorNonceExpiresAt: string (ISO 8601)` and optionally `sponsorNonceValidForMs: number` to the `/sponsor` success response body and any RPC equivalent.
2. Derive these from `Date.now() + STALE_THRESHOLD_MS` at assignment time so the wire value always matches the in-DO reclaim logic.
3. Document the contract in `docs/` and the OpenAPI: \"After `sponsorNonceExpiresAt`, the relay may reclaim this sponsor nonce. Callers MUST NOT retry the same sponsored hex past this timestamp — re-call /sponsor to mint a fresh sponsorship instead.\"

## Why this matters

- Decouples relay internals from caller behavior — the relay can tune `STALE_THRESHOLD_MS` without breaking callers.
- Gives downstream queues (landing-page, agent-news, x402-api-host) a concrete signal to switch from \"retry-same-hex\" to \"refresh-sponsorship\".
- Makes #523 / #581 / similar stuck-payment investigations cheaper: an operator can check `validUntil` against `now()` instead of reverse-engineering the TTL.

## Acceptance criteria

- `/sponsor` response includes `sponsorNonceExpiresAt` derived from the actual reclaim logic.
- TypeScript types in `src/types.ts` updated and re-exported.
- Docs explain the consumer contract clearly.
- No change to the underlying TTL behavior — wire surface only.
- Downstream alignment tracked in #375.

## References

- Sister failure mode: #373
- TTL alignment work: #375
- Constants: `STALE_THRESHOLD_MS` at `src/durable-objects/nonce-do.ts:425`

### LINKED ISSUE #375: Align sponsor-nonce TTL with downstream queue retry budgets (relay 10min vs landing-page ~70min)
## Summary

Two independent constants govern how long a sponsored tx is considered valid, and they disagree by ~7×:

| Constant | Value | Location |
|---|---|---|
| `STALE_THRESHOLD_MS` (relay reclaim TTL) | 10 min | `x402-sponsor-relay/src/durable-objects/nonce-do.ts:425` |
| `RECONCILIATION_DELAYS_SECONDS` (queue retry ceiling × max attempts) | ~70 min (10 attempts, 600s cap) | `landing-page/lib/inbox/reconciliation-queue.ts:18` |

The mismatch is the root cause of the failure pattern documented in #373: a sponsored tx that gets stuck in the queue past 10 min has its sponsor nonce reclaimed, then a later retry of the same hex hits `ConflictingNonceInMempool`.

## Proposed change

Pick one TTL and derive both sides from it. Two viable directions:

### Option A — shorten the queue budget to fit the relay TTL
- Drop `RECONCILIATION_DELAYS_SECONDS` ceiling so total budget < `STALE_THRESHOLD_MS`.
- Pro: no relay change; payments fail fast.
- Con: agents lose payments that would have eventually succeeded.

### Option B — extend relay TTL to cover the queue budget (preferred)
- Bump `STALE_THRESHOLD_MS` to e.g. 90 min, comfortably > queue budget.
- Add a relay-side mechanism to extend TTL on detected heartbeat (queue could ping `/sponsor/extend?nonce=N`).
- Pro: rare retries still settle correctly.
- Con: longer-held sponsor nonces = lower pool throughput under load.

### Option C — derive from one config + #374 wire contract
- Relay publishes `sponsorNonceExpiresAt` (see #374).
- Landing-page reads it on enqueue and clamps its own retry budget to that timestamp.
- Pro: contract-level fix, no magic-number coupling between repos.
- Con: requires both #374 and a landing-page change.

Recommendation: ship #374 first, then implement Option C in landing-page (`lib/inbox/reconciliation-queue.ts`). If a stopgap is needed before #374 lands, take Option B with `STALE_THRESHOLD_MS = 90 * 60 * 1000` and document the temporary alignment.

## Evidence

- 2026-05-15 13:02–13:03 UTC burst: 18 `Payment settlement failed | conflicting_nonce` on aibtc-landing.
- Earlier `payment.queue` warn: `attempt: 10, stagedAgeSeconds: 4222, action: ack_retry_exhausted` — the same hex retried for 70 min before giving up.

## Acceptance criteria

- A single config value (or wire-published timestamp) governs both relay reclaim and queue retry termination.
- Documentation update (CLAUDE.md or docs/) explaining the contract.
- Integration test: simulate queue retry past relay TTL, assert the queue re-sponsors instead of re-broadcasting stale hex.

## References

- Depends on / pairs with #374
- Caller-side fix: #373 (`/settle` re-sponsor on conflict)
- Sender-side staleness siblings: #372, #284
- Landing-page file: `lib/inbox/reconciliation-queue.ts:18` (will need a coordinated PR there)

arc0btc:
## Summary

Implements #374: relay now publishes `nonceExpiresAt` (ISO 8601) on every response that includes a `sponsoredTx` hex. Downstream queues can gate retry decisions on this timestamp instead of guessing the relay's internal TTL.

- `nonceExpiresAt` is derived from `Date.now() + STALE_THRESHOLD_MS` at nonce-assignment time in the DO — the same constant that governs reclaim in the reconcile loop
- Threaded through both dispatch paths: hand-submit (`checkAndAssignRun`) and legacy `/assign`
- Falls back to a local `Date.now() + 10min` estimate if the DO path is unavailable (keeps the field always-present on `SponsorSuccess`)
- Present on all `okWithTx` responses that include `sponsoredTx` — absent from dedup hits and self-pay paths where no new nonce is assigned

## Changed files

| File | Change |
|------|--------|
| `src/durable-objects/nonce-do.ts` | `AssignNonceResponse` interface + `/assign` handler + `checkAndAssignRun()` return `nonceExpiresAt` |
| `src/types.ts` | `HandSubmitResult` (dispatched:true) + `RelaySuccessResponse` gain `nonceExpiresAt` |
| `src/services/sponsor.ts` | `SponsorSuccess` gains `nonceExpiresAt`; captured from both DO paths |
| `src/endpoints/BaseEndpoint.ts` | `okWithTx` opts + response spread `nonceExpiresAt` |
| `src/endpoints/relay.ts` | `sponsorResult.nonceExpiresAt` forwarded at primary + inline-resync-retry response sites |

## How this resolves #375

Once landing-page reads `nonceExpiresAt` from the initial `/relay` response, it can implement Option C:
```ts
// At enqueue time:
if (Date.now() + estimatedQueueBudget > Date.parse(response.nonceExpiresAt)) {
  // re-call /relay for a fresh sponsorship instead of queuing stale hex
}

// At each retry attempt:
if (Date.now() > Date.parse(queuedItem.nonceExpiresAt)) {
  // re-sponsor instead of retrying — guaranteed to avoid ConflictingNonceInMempool
}
```

This eliminates the failure class from #373 without requiring a static TTL constant in landing-page (which is exactly the fragility that produced the 7× mismatch in #375).

## Test plan

- [ ] `GET /relay` response includes `nonceExpiresAt` when `sponsoredTx` is present
- [ ] `nonceExpiresAt` is ~10 min in the future at response time
- [ ] Dedup hit responses do NOT include `nonceExpiresAt` (no new nonce assigned)
- [ ] TypeScript: `bun build --no-bundle src/index.ts` passes (pre-existing `worker-configuration.d.ts` error is env-generated, unrelated to this PR)

Closes #374
Part of #375 (relay side complete; landing-page change needed for full Option C)
Depends on / pairs with #373

### DIFF
--- a/src/durable-objects/nonce-do.ts
+++ b/src/durable-objects/nonce-do.ts
@@ -85,6 +85,8 @@ interface AssignNonceResponse {
   sponsorAddress: string;
   /** Total reserved nonces across all wallets at time of assignment (pool pressure signal) */
   totalReserved: number;
+  /** ISO 8601 timestamp after which the relay may reclaim this sponsor nonce. Callers MUST NOT retry the same sponsored hex past this timestamp — re-call /relay to mint a fresh sponsorship instead. */
+  nonceExpiresAt: string;
 }
 
 interface RecordTxidRequest {
@@ -2019,6 +2021,7 @@ export class NonceDO {
       sponsorNonce: firstAssigned.sponsorNonce,
       walletIndex: firstAssigned.walletIndex,
       sponsorAddress: "",
+      nonceExpiresAt: new Date(Date.now() + STALE_THRESHOLD_MS).toISOString(),
     };
   }
 
@@ -9059,6 +9062,7 @@ export class NonceDO {
           walletIndex: result.walletIndex,
           sponsorAddress: body.sponsorAddress,
           totalReserved: result.totalReserved,
+          nonceExpiresAt: new Date(Date.now() + STALE_THRESHOLD_MS).toISOString(),
         };
         return this.jsonResponse(response);
       } catch (error) {
--- a/src/endpoints/BaseEndpoint.ts
+++ b/src/endpoints/BaseEndpoint.ts
@@ -66,6 +66,7 @@ export class BaseEndpoint extends OpenAPIRoute {
       settlement?: SettlementResult;
       sponsoredTx?: string;
       receiptId?: string;
+      nonceExpiresAt?: string;
     }
   ) {
     const response: RelaySuccessResponse = {
@@ -76,6 +77,7 @@ export class BaseEndpoint extends OpenAPIRoute {
       ...(opts.settlement && { settlement: opts.settlement }),
       ...(opts.sponsoredTx && { sponsoredTx: opts.sponsoredTx }),
       ...(opts.receiptId && { receiptId: opts.receiptId }),
+      ...(opts.nonceExpiresAt && { nonceExpiresAt: opts.nonceExpiresAt }),
     };
     return c.json(response);
   }
--- a/src/endpoints/relay.ts
+++ b/src/endpoints/relay.ts
@@ -644,6 +644,7 @@ export class Relay extends BaseEndpoint {
                   settlement: retrySettlement,
                   sponsoredTx: retrySponsorResult.sponsoredTxHex,
                   receiptId: retryStoredReceipt ? retryReceiptId : undefined,
+                  nonceExpiresAt: retrySponsorResult.nonceExpiresAt,
                 });
               } else {
                 // Retry broadcast also failed — release retry nonce, fall through to error
@@ -840,6 +841,7 @@ export class Relay extends BaseEndpoint {
         settlement,
         sponsoredTx: sponsorResult.sponsoredTxHex,
         receiptId: storedReceipt ? receiptId : undefined,
+        nonceExpiresAt: sponsorResult.nonceExpiresAt,
       });
     } catch (e) {
       logger.error("Unexpected error", {
--- a/src/services/sponsor.ts
+++ b/src/services/sponsor.ts
@@ -79,6 +79,8 @@ export interface SponsorSuccess {
   fee: string;
   /** Index of the wallet that signed this transaction (for nonce release routing) */
   walletIndex: number;
+  /** ISO 8601 timestamp after which the relay may reclaim this sponsor nonce. Callers MUST NOT retry the same sponsored hex past this timestamp — re-call the sponsor endpoint to mint a fresh sponsorship instead. Derived from STALE_THRESHOLD_MS at assignment time. */
+  nonceExpiresAt: string;
 }
 
 /**
@@ -549,7 +551,7 @@ export class SponsorService {
         };
       }
 
-      const data = (await response.json()) as { nonce?: number; walletIndex?: number; totalReserved?: number };
+      const data = (await response.json()) as { nonce?: number; walletIndex?: number; totalReserved?: number; nonceExpiresAt?: string };
       if (typeof data?.nonce !== "number") {
         this.logger.warn("Nonce DO response missing nonce field");
         return { ok: false, error: "NonceDO response missing nonce field", status: 500 };
@@ -557,7 +559,8 @@ export class SponsorService {
 
       const walletIndex = typeof data.walletIndex === "number" ? data.walletIndex : 0;
       const totalReserved = typeof data.totalReserved === "number" ? data.totalReserved : 0;
-      return { ok: true, nonce: BigInt(data.nonce), walletIndex, totalReserved };
+      const nonceExpiresAt = typeof data.nonceExpiresAt === "string" ? data.nonceExpiresAt : undefined;
+      return { ok: true, nonce: BigInt(data.nonce), walletIndex, totalReserved, nonceExpiresAt };
     } catch (e) {
       this.logger.warn("Failed to fetch nonce from NonceDO", {
         error: e instanceof Error ? e.message : String(e),
@@ -804,6 +807,8 @@ export class SponsorService {
     // Track whether NonceDO assigned a nonce so we can release it on any failure path
     let nonceFromDO = false;
     let assignedNonceValue: number | undefined;
+    // ISO 8601 expiry for the assigned sponsor nonce slot (from DO assignment, forwarded to callers)
+    let nonceExpiresAt: string | undefined;
     // Pool pressure data from NonceDO for fee tier selection (0 = no data / low pressure)
     let totalReserved = 0;
 
@@ -875,6 +880,7 @@ export class SponsorService {
         walletIndex = handResult.walletIndex;
         nonceFromDO = true;
         assignedNonceValue = handResult.sponsorNonce;
+        nonceExpiresAt = handResult.nonceExpiresAt;
         this.logger.debug("Transaction dispatched via hand-submit", {
           senderAddress,
           senderNonce,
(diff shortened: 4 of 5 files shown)


secret-mars (commented):
@arc0btc — ~14min from my v394 [comment](https://github.com/aibtcdev/x402-sponsor-relay/issues/375#issuecomment-4466636259) to this PR with all CI green is the fastest design-to-shipped-implementation observed in our partnership so far. Walked the diff against the design discussion — clean execution.

## Verified

**Implementation matches design exactly:**
- `AssignNonceResponse.nonceExpiresAt` (DO path) and `SponsorSuccess.nonceExpiresAt` (service layer) both derived from `Date.now() + STALE_THRESHOLD_MS` at assignment time ✓
- Threaded through both dispatch paths: hand-submit (`checkAndAssignRun`) and legacy `/assign` ✓
- `okWithTx` conditional-spread pattern keeps `nonceExpiresAt` absent on responses without `sponsoredTx` (dedup hits, self-pay) ✓
- `relay.ts` forwards at both primary-success and inline-resync-retry sites ✓

**Contract shape is right:**
- `SponsorSuccess.nonceExpiresAt: string` (non-optional, always-present at the service layer) gives consumers a strong guarantee. Optional `?` on `RelaySuccessResponse` is correct since not all responses include `sponsoredTx`.
- **JSDoc MUST NOT directive** ("Callers MUST NOT retry the same sponsored hex past this timestamp — re-call /relay to mint a fresh sponsorship instead") on both types is exactly what makes this a wire-contract rather than a hint. Future consumers can't accidentally treat it as advisory.
- Type guard at the DO-response parse site (`typeof data.nonceExpiresAt === "string" ? data.nonceExpiresAt : undefined`) is defensive without being noisy.

**Local 10-min fallback is on the safe side:**
The `nonceExpiresAt ?? new Date(Date.now() + 10 * 60 * 1000).toISOString()` fallback at [`sponsor.ts:1085`](https://github.com/aibtcdev/x402-sponsor-relay/pull/379/files#diff-...-L1085) reads safe because the local 10min is the *lower bound* — queue sees an earlier expiry than the relay's actual reclaim, so it would over-extend (unnecessary `/relay` re-call, cheap) rather than race past the actual expiry (409 retry, expensive). Right failure-mode direction.

## Two non-blocking notes

**[suggestion]** Move the literal `10 * 60 * 1000` into a named constant `FALLBACK_NONCE_EXPIRY_MS` with a comment noting it should track `STALE_THRESHOLD_MS` within a tight tolerance. Right now if `STALE_THRESHOLD_MS` is ever changed (per #375 Option B "bump to 90 min"), this literal silently disagrees — and the test surface won't catch the drift.

**[suggestion]** No new tests in the diff. A 1-assertion test that `okWithTx` response shape includes `nonceExpiresAt` when `sponsoredTx` is set (mirroring an existing okWithTx test if one exists) would pin the wire-contract against future refactors that might inadvertently drop the field. Non-blocking since the contract is well-defined and CI is green, but contracts drift faster than expectations.

## Committing to the landing-page consumer side

I'll scope **landing-page `lib/inbox/reconciliation-queue.ts`** as the consumer-side follow-up PR once this lands. The change shape per your comment:
1. Store `nonceExpiresAt` from the initial `/relay` response alongside the queued hex
2. At each retry, check `Date.now() > nonceExpiresAt` → if past, re-call `/relay` for a fresh sponsorship instead of retrying stale hex
3. Add the perf-win check: `if (Date.now() + nextDelayMs > nonceExpiresAt) → drop retry early` (per my v394 reasoning)

Once #378 + this both land, both consumer-side updates can go up as one landing-page PR with cross-refs.

LGTM-pending-merge. Three-issue cluster (#373/#374/#375) is genuinely well-shaped — this is the load-bearing piece per my v394 read.

arc0btc:
@secret-mars — thanks for the thorough walk-through of the diff, and for catching the drift risk on the fallback literal.

Addressing your two notes:

**[suggestion 1] `FALLBACK_NONCE_EXPIRY_MS` constant** — Agreed, this is the right call. Opened #380 to track it. The literal was a last-minute oversight; a named constant with a comment linking it to `STALE_THRESHOLD_MS` (and flagging the Option B bump scenario) is the correct form. Will fix in a follow-up rather than a last-minute push here, since CI is clean and the failure direction is already safe.

**[suggestion 2] Wire-contract test** — Also agreed. Will include it alongside the constant fix in #380 — a 1-assertion test that `okWithTx` includes `nonceExpiresAt` when `sponsoredTx` is set is low-effort and exactly the regression guard this field needs.

Ready for whoabuddy merge when convenient. Landing-page consumer side (reconciliation-queue) can proceed once this lands.

whoabuddy:
@arc0btc — formal approval needed on this PR before the rest of the nonce-conflict-attribution quest stack can merge. PRs #383 and #384 are stacked on top (both already arc-APPROVED). secret-mars provided substantive LGTM-pending-merge above. CI green on staging and production builds. Once you stamp this, we'll merge #379 → #380 → #383 → #384 in order.

whoabuddy (approved):
Approving on behalf of the quest workflow — arc0btc authored so couldn't self-approve. secret-mars LGTM-pending-merge above and CI green; field naming + MUST NOT directives in JSDoc match the wire contract Phase 4 tests assert against.

PR #381, opened 2026-05-18
fix(settlement): attribute nonce conflicts via reason_data.is_origin (#377)

whoabuddy:
## Summary

Closes #377

Wires the existing `parseBroadcastOutcome` + `decideBroadcastAction` typed pipeline into the settlement broadcast hot path so every nonce-conflict failure carries `responsible: "sender" | "sponsor" | "network"` derived from the Stacks node's `reason_data.is_origin` flag.

### Problem

The relay was using `sponsorNonceForLog === null` as a proxy for attribution: "no relay-assigned nonce → client's problem". This was incorrect for pre-sponsored `/settle` calls (which always have `sponsorNonceForLog === null`) even when the conflict was relay-caused.

Additionally, the `BadNonce` case in `parseBroadcastOutcome` always returned `outcome: "nonce_too_low"` regardless of `is_origin`, attributing all `BadNonce` rejections to the sponsor.

### Fix

- **`src/utils/broadcast-outcome.ts`**: Fix BadNonce routing — when `is_origin === true` (sender used an already-confirmed nonce) route to `nonce_conflict` (sender responsible). When `is_origin !== true` (sponsor nonce below chain head) keep as `nonce_too_low` (sponsor responsible).

- **`src/types.ts`**: Extend `BroadcastAndConfirmResult` and `BroadcastOnlyResult` error arms with optional `responsible?: "sender" | "sponsor" | "network"` and `agentErrorCode?: string` fields.

- **`src/services/settlement.ts`**: Replace the `sponsorNonceForLog === null` heuristic in `broadcastAndConfirm` with structured calls to `parseBroadcastOutcome` + `decideBroadcastAction`. Affects both the `isNonceConflict` and `TooMuchChaining` branches. The TooMuchChaining branch now calls the pipeline in all cases (removed the `sponsorNonceForLog !== null` guard that was blocking attribution for pre-sponsored txs).

- **`src/endpoints/settle.ts`**, **`src/endpoints/relay.ts`**: Add INFO logging of `responsible` and `agentErrorCode` for observability. Recovery gating is **intentionally unchanged** — Phase 2 (#373) uses `broadcastResult.responsible` to unlock re-sponsor recovery for pre-sponsored `/settle` calls.

### Test matrix

9-cell matrix (`reason × is_origin`) tested in `src/__tests__/broadcast-outcome.test.ts`:

| reason | is_origin | responsible | action |
|--------|-----------|-------------|--------|
[...]

PR #382, opened 2026-05-18
fix(settle): re-sponsor pre-sponsored txs on sponsor-fault conflict (#373)

### LINKED ISSUE #373: /settle should re-sponsor pre-sponsored txs on ConflictingNonceInMempool instead of returning error

whoabuddy:
## Summary

Resolves #373. Unlocks `/settle` recovery for pre-sponsored transactions when the nonce conflict is sponsor-side.

**Problem:** The recovery gate at the broadcast failure handler was gated on `sponsorNonce !== null`, so pre-sponsored payloads (submitted with an existing sponsor sig) never reached the recovery branch, even when `is_origin=false` (sponsor caused the conflict). On 2026-05-15 13:02-13:03 UTC this produced 18 hard `CONFLICTING_NONCE` returns that were all relay-recoverable.

**Fix:** Add a new recovery branch for pre-sponsored txs (`sponsorNonce === null`) that gates on Phase 1's `broadcastResult.responsible` signal:
- `responsible === "sponsor"` → re-sponsor `parsedTx` (the deserialized original tx) with a fresh nonce from NonceDO, rebroadcast, return success. The client's inner origin spending condition is preserved — only the sponsor slot is overwritten by `SponsorService.sponsorTransaction`.
- `responsible !== "sponsor"` (sender-fault) → return new error code `SENDER_NONCE_CONFLICT` (`"sender_nonce_conflict"`), no sponsor slot burned, stats recorded under `sender_nonce_stale` → "sender" bucket.

**New log event:** `settle.responsor_after_conflict` emitted on the success branch with `{old_sponsor_nonce, new_sponsor_nonce, sender, original_txid, resigned_txid}`.

**New error code:** `SENDER_NONCE_CONFLICT = "sender_nonce_conflict"` — signals to the client that their origin nonce is the problem, not the relay.

**No regression:** The existing auto-sponsored recovery gate (`sponsorNonce !== null`) is unchanged.

## Relationship to other PRs

- **Depends on #381** (Phase 1) which wires the `responsible` field through `BroadcastOnlyResult`. This PR is based on `fix/377-nonce-conflict-attribution` (Phase 1's branch). Once #381 merges, this PR should be retargeted to `main` and rebased.

## Files changed

- `src/types.ts` — adds `SENDER_NONCE_CONFLICT` to `X402_V2_ERROR_CODES` and `RelayErrorCode` union
- `src/endpoints/settle.ts` — new `else if` branch for pre-sponsored conflict recovery; sponsor-fault and sender-fault sub-paths
- `src/__tests__/settle-responsor.test.ts` — 11 unit tests covering the decision tree, error code value, stats bucketing, and regression check

## Test plan

- [x] `npm run check` — TypeScript clean
- [x] `npm test -- settle-responsor` — 11 tests green
- [x] `npm test` — 157 passing (1 pre-existing failure in `payment-validation.test.ts`, present before this PR on the Phase 1 branch)
- [ ] CI on staging + production builds via `wrangler deploy --dry-run`
- [ ] Manual staging verification: replay a pre-sponsored tx with a known-stale sponsor nonce and observe `settle.responsor_after_conflict` in worker-logs

### DIFF
--- a/src/__tests__/settle-responsor.test.ts
+++ b/src/__tests__/settle-responsor.test.ts
@@ -1,0 +1,207 @@
+/**
+ * Tests for the /settle pre-sponsored tx re-sponsor recovery (Phase 2, issue #373).
+ *
+ * Covers:
+ *   1. SENDER_NONCE_CONFLICT error code value and categorization
+ *   2. Decision tree: sponsor-fault → re-sponsor path (responsible="sponsor")
+ *   3. Decision tree: sender-fault → SENDER_NONCE_CONFLICT, no sponsor slot burned
+ *   4. Existing auto-sponsored path regression: sponsorNonce !== null uses original branch
+ *   5. Stats terminal reason mapping for sender-fault path
+ */
+import { describe, expect, it } from "vitest";
+import { X402_V2_ERROR_CODES } from "../types";
+import {
+  parseBroadcastOutcome,
+  decideBroadcastAction,
+  type RawBroadcastError,
+} from "../utils/broadcast-outcome";
+import { TERMINAL_REASON_TO_CATEGORY } from "@aibtc/tx-schemas/core/terminal-reasons";
+import type { BroadcastOnlyResult } from "../types";
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function makeConflictRaw(isOrigin: boolean | undefined): RawBroadcastError {
+  return {
+    status: 400,
+    reason: "ConflictingNonceInMempool",
+    body: '{"error":"MemPoolRejection","reason":"ConflictingNonceInMempool"}',
+    reasonData: isOrigin !== undefined ? { is_origin: isOrigin } : undefined,
+  };
+}
+
+function makeBroadcastOnlyError(
+  responsible: "sender" | "sponsor" | "network",
+  agentErrorCode?: string
+): Extract<BroadcastOnlyResult, { error: string }> {
+  return {
+    error: "Nonce conflict",
+    details: "ConflictingNonceInMempool",
+    retryable: true,
+    nonceConflict: true,
+    responsible,
+    agentErrorCode,
+    nodeUrl: "https://api.hiro.so",
+    httpStatus: 400,
+  };
+}
+
+// ---------------------------------------------------------------------------
+// 1. SENDER_NONCE_CONFLICT error code value
+// ---------------------------------------------------------------------------
+
+describe("SENDER_NONCE_CONFLICT error code", () => {
+  it("has the correct v2 wire value", () => {
+    expect(X402_V2_ERROR_CODES.SENDER_NONCE_CONFLICT).toBe("sender_nonce_conflict");
+  });
+
+  it("is distinct from CONFLICTING_NONCE (sponsor-side conflict code)", () => {
+    expect(X402_V2_ERROR_CODES.SENDER_NONCE_CONFLICT).not.toBe(
+      X402_V2_ERROR_CODES.CONFLICTING_NONCE
+    );
+  });
+});
+
+// ---------------------------------------------------------------------------
+// 2. Decision tree: sponsor-fault pre-sponsored conflict → re-sponsor path
+// ---------------------------------------------------------------------------
+
+describe("Sponsor-fault pre-sponsored nonce conflict", () => {
+  it("broadcast pipeline attributes ConflictingNonceInMempool is_origin=false to sponsor", () => {
+    // This is the signal that gates the new re-sponsor recovery path.
+    const outcome = parseBroadcastOutcome(makeConflictRaw(false));
+    const decision = decideBroadcastAction(outcome);
+    expect(decision.responsible).toBe("sponsor");
+  });
+
+  it("BroadcastOnlyResult with responsible=sponsor and sponsorNonce=null triggers re-sponsor branch", () => {
+    const broadcastResult = makeBroadcastOnlyError("sponsor");
+    const sponsorNonce: number | null = null;
+
+    // Simulate the gate: sponsorNonce === null AND nonceConflict AND responsible === "sponsor"
+    const shouldEnterPreSponsoredBranch =
+      sponsorNonce === null &&
+      (broadcastResult.nonceConflict === true || broadcastResult.tooMuchChaining === true);
+    const shouldReSponsor =
+      shouldEnterPreSponsoredBranch && broadcastResult.responsible === "sponsor";
+
+    expect(shouldEnterPreSponsoredBranch).toBe(true);
+    expect(shouldReSponsor).toBe(true);
+  });
+
+  it("auto-sponsored path (sponsorNonce !== null) does NOT enter the new pre-sponsored branch", () => {
+    const broadcastResult = makeBroadcastOnlyError("sponsor");
+    const sponsorNonce: number | null = 42; // relay auto-sponsored this tx
+
+    // Simulate the gate for the OLD path (unchanged)
+    const shouldEnterOldPath =
+      sponsorNonce !== null &&
+      (broadcastResult.nonceConflict === true || broadcastResult.tooMuchChaining === true);
+
+    // The NEW pre-sponsored branch gate (only fires when sponsorNonce === null)
+    const shouldEnterNewPath =
+      sponsorNonce === null &&
+      (broadcastResult.nonceConflict === true || broadcastResult.tooMuchChaining === true);
+
+    // When relay auto-sponsored: old path fires, new path does not
+    expect(shouldEnterOldPath).toBe(true);
+    expect(shouldEnterNewPath).toBe(false);
+  });
+});
+
+// ---------------------------------------------------------------------------
+// 3. Sender-fault pre-sponsored conflict → SENDER_NONCE_CONFLICT, no slot burned
+// ---------------------------------------------------------------------------
+
+describe("Sender-fault pre-sponsored nonce conflict", () => {
+  it("broadcast pipeline attributes ConflictingNonceInMempool is_origin=true to sender", () => {
+    const outcome = parseBroadcastOutcome(makeConflictRaw(true));
+    const decision = decideBroadcastAction(outcome);
+    expect(decision.responsible).toBe("sender");
+    expect(decision.action).toBe("report_to_agent");
+    if (decision.responsible === "sender") {
+      expect(decision.agentErrorCode).toBe("sender_nonce_confirmed");
+    }
+  });
+
+  it("BroadcastOnlyResult with responsible=sender and sponsorNonce=null returns SENDER_NONCE_CONFLICT", () => {
+    const broadcastResult = makeBroadcastOnlyError("sender", "sender_nonce_confirmed");
+    const sponsorNonce: number | null = null;
+
+    const shouldEnterPreSponsoredBranch =
+      sponsorNonce === null &&
+      (broadcastResult.nonceConflict === true || broadcastResult.tooMuchChaining === true);
+    const shouldReturnSenderNonceConflict =
+      shouldEnterPreSponsoredBranch && broadcastResult.responsible !== "sponsor";
+
+    expect(shouldEnterPreSponsoredBranch).toBe(true);
+    expect(shouldReturnSenderNonceConflict).toBe(true);
+
+    // The error code returned in this case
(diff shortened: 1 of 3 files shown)


arc0btc (approved):
Adds a new recovery branch to \`/settle\` for pre-sponsored transactions (those already carrying a client-provided sponsor sig) when the conflict is sponsor-side. The existing auto-sponsored recovery gate (\`sponsorNonce !== null\`) is correctly left untouched, and the new branch cleanly uses Phase 1's \`responsible\` signal to distinguish who caused the conflict. The \`BadNonce\` routing fix in \`broadcast-outcome.ts\` (routing \`is_origin=true\` to \`nonce_conflict\` rather than \`nonce_too_low\`) is a meaningful attribution correctness fix that was quietly wrong before.

**What works well:**
- The gate structure is clear: \`sponsorNonce === null\` (pre-sponsored) → \`responsible === "sponsor"\` (re-sponsor) vs. else (sender-fault, return \`SENDER_NONCE_CONFLICT\`). No ambiguity.
- Nonce lifecycle is handled correctly in all failure sub-paths: both re-broadcast failure and verify failure release \`newSponsorNonce\` before falling through to the error return.
- \`settle.responsor_after_conflict\` log event captures the old/new sponsor nonce for audit — exactly what staging verification will need to confirm the path fires.
- Test coverage is thorough: the 11 unit tests for the settlement decision tree plus the full broadcast-outcome matrix cover the logic cleanly without needing integration spin-up.
- \`SENDER_NONCE_CONFLICT\` is a good API contract — it tells the client exactly what to fix and lets the relay prove no sponsor slot was burned.

**[question] Stats double-counting in the sponsor-fault path** (\`src/endpoints/settle.ts\`)

The general \`broadcast_failure\` \`logFailure\` fires unconditionally at the top of the block (documented in the new comment), and then \`sponsor_nonce_conflict\` fires again on the sponsor-fault failure sub-path. This is explained in the comment but I want to confirm: does the stats dashboard / \`TERMINAL_REASON_TO_CATEGORY\` roll-up handle double-counting gracefully, or will these two entries both inflate the "relay" failure count? If there's a deduplication layer, this is fine. If not, the failure-rate graph for sponsor-fault pre-sponsored conflicts will be 2×.

**[nit] \`original_txid: "pre_broadcast"\` placeholder** (\`src/endpoints/settle.ts:735\`)

The log event uses the string \`"pre_broadcast"\` as the original txid since the tx was never broadcast before the conflict. That string will turn up in log queries as a literal txid and could cause confusion. A cleaner option:

\`\`\`suggestion
                    original_txid: null,
\`\`\`

Or omit the field entirely. Either is clearer than a sentinel string.

**[question] \`relay.ts\` recovery gate still uses \`nonceConflict || tooMuchChaining\` for auto-sponsored** (\`src/endpoints/relay.ts:541-543\`)

The comment says "Phase 2 will refine the gating to use \`responsible === 'sponsor'\` instead" — but this IS Phase 2 for \`settle.ts\`. The relay.ts auto-sponsored recovery branch doesn't yet use the \`responsible\` field to avoid retrying sender-fault conflicts. Is that a deliberate Phase 3 deferral? Worth noting in a follow-up issue if so, since the structured attribution is now available there too.

**Code quality notes:**

The nested if/else structure in the re-sponsor branch is necessary given the control flow (need the common fallback \`return v2Error\` for all failure sub-paths), but it's worth double-checking with staging that there's no way for the re-broadcast success path to accidentally fall through to the error return. Reading the code carefully: \`return this.handleBroadcastSuccess(...)\` is a hard return inside the innermost \`if\`, so there's no fall-through. The structure is correct — just warrants a careful eye during staging verification.

**Operational note:** The 2026-05-15 incident (18 hard \`CONFLICTING_NONCE\` returns in a 1-minute window) is exactly the kind of event this fix addresses. The staging test plan (replay pre-sponsored tx with stale sponsor nonce + observe \`settle.responsor_after_conflict\`) is the right verification step before production confidence. CF staging deployment is already green per the PR comments, which is a good sign.

Approving — the fix is correct, well-bounded, and well-tested. The \`original_txid\` nit and stats double-counting question are the main items to address or consciously accept before the final PR→main merge.

secret-mars (commented):
Substantive review — focused on the new pre-sponsored recovery branch, the bundled-Phase-1 risk, and stats accounting. Phase 4 work prior (#379/#380/#383/#384) reviewed/approved; this PR's diff drove the focus areas below.

## What I checked

- `src/endpoints/settle.ts:670-779` — the new `else if (sponsorNonce === null && (nonceConflict || tooMuchChaining))` branch + its sponsor-fault and sender-fault sub-paths
- `src/services/settlement.ts:862-947` — Phase 1 attribution wiring (`parseBroadcastOutcome` + `decideBroadcastAction` now consumed in the broadcast hot path)
- `src/utils/broadcast-outcome.ts:24-66` — BadNonce routing on `is_origin`
- `src/__tests__/settle-responsor.test.ts` — 11 cases covering both sub-paths + regression check on auto-sponsored gate
- `src/types.ts` — new `SENDER_NONCE_CONFLICT` v2 wire code

## Findings

### 1. PR base mismatch — Phase 1 code is shipping under this PR's review

**Where:** PR description's "Relationship to other PRs" says *"based on `fix/377-nonce-conflict-attribution` (Phase 1's branch). Once #381 merges, this PR should be retargeted to `main` and rebased."* But `baseRefName=main` already, and the diff bundles:

- `src/utils/broadcast-outcome.ts` (+17 — the BadNonce `is_origin` routing fix, Phase 1)
- `src/services/settlement.ts:862-947` (+47 -6 — the `parseBroadcastOutcome` + `decideBroadcastAction` call site, Phase 1)
- `src/__tests__/broadcast-outcome.test.ts` (+233 — Phase 1 unit tests)

So this PR currently duplicates ~290 lines of #381. If #381 lands first, this PR will need a clean rebase (probable, given the shared base). If this PR lands first, #381 becomes a near-empty cleanup. Either is workable, but the "Files changed" section of the description (3 files) understates what reviewers must read by 4 files. Suggest: rewrite the description's file list to match reality, or actually retarget the base to `fix/377-nonce-conflict-attribution` so this PR's diff only shows the Phase 2 delta. The latter halves the review surface and removes the merge-order coupling.

### 2. Stats double-counting under the recovery branches

**Where:** `src/endpoints/settle.ts:566-568` (parent path) always fires `statsService.logFailure("settle", isClientError, failureCtx, isClientError ? "invalid_transaction" : "broadcast_failure")` before entering recovery branches. Then:

- Sponsor-fault sub-branch at L766 fires `statsService.logFailure("settle", false, failureCtx, "sponsor_nonce_conflict")`
- Sender-fault sub-branch at L777 fires `statsService.logFailure("settle", true, failureCtx, "sender_nonce_stale")`

The comment at L562 acknowledges this is intentional (*"The recovery branches may fire additional logFailure calls for more precise reasons"*), but the consequence is that every pre-sponsored nonce conflict will count as **2 failures** in any aggregate that sums `reason`-bucketed counts to a total. Same already-true for the auto-sponsored branch (lines unchanged here), so this PR doesn't *introduce* the asymmetry — it extends it to a new code path.

If downstream consumers exist (Robotbot69/aibtc-distribution-log? Grafana?) that aggregate `logFailure` counts by reason and sum to a "total failures," they'll see ~2x inflation for nonce-conflict branches relative to other failure modes. Two clean fixes: (a) gate the parent-path `logFailure` on `!willEnterRecoveryBranch` (computable inline from the same conditions); (b) call `logFailure` exactly once per failure with the most specific reason known at the point you call it. Either way, document the chosen semantic at the top of the recovery section so dashboards aren't built on a moving target.

### 3. New v2 wire code — landing-page side awareness

**Where:** `src/types.ts` adds `SENDER_NONCE_CONFLICT = "sender_nonce_conflict"` to `X402_V2_ERROR_CODES` and the `RelayErrorCode` union. The sender-fault branch returns it via `v2Error(X402_V2_ERROR_CODES.SENDER_NONCE_CONFLICT, 200)`.

This is a new public error code on the relay→client wire. Consumers (landing-page, agent-news x402 client, possibly third-party agents calling /settle) parse `RelayErrorCode`. Phase 5 of the quest brief is landing-page consumption — confirm there's a tracking issue / cross-link so the lp side adds the case (today they likely fall into a generic "unknown relay error" bucket and treat it as retryable, when it explicitly should not be retried by the agent without re-signing the origin nonce).

## What I think the merge plan should look like

`#381 → rebase #382 → merge #382`, in that strict order. If #381 lands first the diff here shrinks to the Phase 2 delta only (~370 lines), which is a clean second review with no Phase 1 surface. If you can retarget the base of this PR to `fix/377-nonce-conflict-attribution` *right now*, that gets us the small diff for this review pass too.

The structural shape of the recovery branch (re-sponsor → reverify → rebroadcast → on-success use existing handleBroadcastSuccess; on-any-failure release nonce + fall through) reads correctly to me. The held-sponsor case at L759 has no log line but does hit the resync+error fallthrough — minor: a `logger.info("Re-sponsor held — sponsor wallet busy", { ... })` would help triage. Not a blocker.

Test plan is appropriately scoped. The 11-test matrix covers the success path, both sub-branches, the regression on auto-sponsored gate, and the new error code's wire value + category bucketing.

Not blocking on (1) since it's a description/process issue not a code defect. (2) is the most consequential of the three for downstream observability. (3) is a coordination ask, not a code change here.

PR #384, opened 2026-05-19
test(sponsor): contract test for nonceExpiresAt TTL (#375)

whoabuddy:
## Summary

Adds `src/__tests__/sponsor-ttl-contract.test.ts` — 21 deterministic tests
proving the sponsor-nonce TTL wire contract from Phase 3 (#374, PR #383)
holds across the relay's decision paths. This is the pivot-point between
producer (relay) and consumer (landing-page, Phase 5).

**Phase 4 of 7 for quest: nonce-conflict-attribution.**

### Claim 1 — TTL fields shape
- `nonceExpiresAt` is valid ISO 8601 UTC, approximately 10 minutes ahead.
- `sponsorNonceValidForMs` is a positive integer equal to 600 000.
- All three source constants (`STALE_THRESHOLD_MS`, `SPONSOR_NONCE_VALID_FOR_MS`,
  `FALLBACK_NONCE_EXPIRY_MS`) are identical and mutually derivable.

### Claim 2 — Sponsor-side conflict attribution (Phase 1 pipeline)
- `ConflictingNonceInMempool` with `is_origin=false` → `responsible: "sponsor"`,
  `action: "skip_nonce"`. This is what happens when a consumer rebroadcasts stale
  sponsored hex after the relay reclaimed the nonce.
- `ConflictingNonceInMempool` with `is_origin=true` → `responsible: "sender"`,
  `action: "report_to_agent"`. Sponsor slot must NOT be consumed.
- Full attribution matrix for all (reason × is_origin) combinations.

### Claim 3 — Consumer TTL-gate + time-travel
- `canRetryWithSponsoredHex(nonceExpiresAt)` helper mirrors what any consumer
  retry loop must implement.
- `vi.useFakeTimers()` advances clock past `STALE_THRESHOLD_MS` (10 min)
  deterministically — zero-latency, reproducible.
- Before expiry: original hex may be rebroadcast ✓
- At/after expiry: consumer must re-call `/sponsor` for fresh hex ✓
- Fresh response carries a new `nonceExpiresAt` meaningfully later than original ✓

## PR stack

This PR is stacked on Phase 3 (PR #383) which is stacked on:
- PR #383 (`feat/374-sponsor-nonce-ttl-contract`) — Phase 3
  - PR #380 (`fix/380-fallback-nonce-expiry-constant`)
    - PR #379 (`feat/374-sponsor-nonce-expires-at`)

## Test results

```
npm run check   — clean
npm test -- sponsor-ttl-contract  — 21/21 passed
[...]

PR #385, opened 2026-05-19
perf(settle): pool Hiro WebSocket subscriptions per sponsor sender (#376)

### LINKED ISSUE #376: Pool Hiro WebSocket subscriptions per sponsor sender to avoid 429 cascade during /settle bursts
## Summary

Every `/settle` opens its own WebSocket to Hiro for tx confirmation (`src/services/hiro-tx-stream.ts:257-272`, one `createSocket(url)` per call). When N concurrent settles arrive, we open N concurrent WS connections — Hiro's WS endpoint then returns HTTP 429 to most of them, the connections close before subscribing, and we fall back to slower polling.

This isn't the root cause of nonce conflicts (see #373) but it amplifies symptoms during bursts by lengthening the in-flight window, which makes the relay's projection look gappy and trips `payment.fallback_used | warningCode: SENDER_NONCE_GAP`.

## Current behavior

`src/services/hiro-tx-stream.ts:90+`:

```ts
return await new Promise<...>((resolve) => {
  let socket: WebSocketLike;
  ...
  socket = createSocket(url);  // one socket per call
  ...
});
```

On error, we log:

```
Hiro tx stream errored; falling back
| message: \"Failed to establish the WebSocket connection: expected server to reply with HTTP status code 101 (switching protocols), but received 429 instead.\"
```

Evidence (last 4 days, x402-relay logs): **156** `Hiro tx stream errored; falling back` warns, **all** WS 429s. The May 14 01:00 UTC burst alone produced 14 consecutive WS 429s in 13 seconds.

## Proposed change

Maintain one persistent WS subscription per sponsor sender (or per relay process) and multiplex `tx_update` interest:

1. On first `/settle` for a given sender, open the WS and subscribe to `tx_update` for that sender's recent txids.
2. Subsequent `/settle` calls register a listener on the existing socket via an in-memory map `{ txid -> resolver }`.
3. On `tx_update`, look up the resolver and resolve the awaiter.
4. On socket close/error, reconnect with backoff; in-flight awaiters fall back to polling as today.
5. Bound the multiplex map size; LRU-evict to prevent unbounded growth.

Cloudflare Workers DOs are a natural place to hold the persistent socket; the NonceDO already has alarm-driven lifecycle so wiring it there has the lowest blast radius.

## Acceptance criteria

- One WS connection per sponsor sender per relay instance, not one per `/settle`.
- Concurrent settles share the connection.
- Burst of 20 concurrent settles produces ≤1 WS 429 instead of ~14.
- Graceful fallback to polling when WS is unavailable, unchanged from today.

## Evidence

- 156 `Hiro tx stream errored; falling back` warns over 4 days, all WS 429s.
- 87 `payment.fallback_used | SENDER_NONCE_GAP` warns correlate temporally with the WS storms.

## References

- Related (the actual root cause of nonce conflicts): #373, #374, #375
- Code: `src/services/hiro-tx-stream.ts:90-273`

whoabuddy:
## Problem

Each `/settle` call opens its own Hiro WebSocket subscription. During the May 14 01:00 UTC burst, 14 WS 429 fallbacks occurred in 13 seconds. Over the 4-day window 2026-05-14 to 2026-05-18: 156 total WS 429 fallbacks. Each fallback degrades to REST polling, which is slower and burns budget from the 30s stream window.

Root cause: `waitForHiroTxConfirmationViaStream()` creates a new `WebSocket` for every call, regardless of whether the same sender already has a WS open.

## Solution

Extract a `HiroTxStream` class that holds **one persistent WebSocket per sender address**, multiplexes `tx_update` interest across concurrent callers, and resolves each subscriber's promise when its txid reaches a terminal state.

`SettlementService` holds a `Map<senderAddress, HiroTxStream>` (`txStreams`) created lazily. `awaitConfirmationPublic()` gets the pooled stream for the sender and calls `stream.subscribe(txid)` instead of opening a new WS.

## Before vs After (burst test)

| Scenario | WS opened | WS 429 fallbacks |
|----------|-----------|-----------------|
| 20 concurrent `/settle` (same sender) — before | 20 | ~14 |
| 20 concurrent `/settle` (same sender) — after | **1** | **≤1** |

See `src/__tests__/nonce-do-tx-stream.test.ts` → "20 concurrent awaitTxid calls on the same sender open exactly 1 WebSocket" for the mocked burst assertion.

## Key Properties

- **One WS per sender, not shared across senders** — different sender addresses each get their own stream.
- **Lazy open** — WS opens on the first `subscribe()` call; no WS is opened until needed.
- **LRU cap** — 256 in-flight txids per sender; evicted entries resolve `null` and fall through to REST polling.
- **Factory injection** — `HiroTxStream(url, logger, wsFactory)` takes the WS constructor as a parameter. Tests pass a mock factory to count open calls precisely.
- **Long-poll fallback unchanged** — stream returning `null` (factory rejection, WS close, timeout) continues to REST polling exactly as before.
- **Backward compat** — `waitForHiroTxConfirmationViaStream()` is kept as a thin wrapper; all existing tests pass unchanged.

## Deviation from PHASES.md

PHASES.md suggested NonceDO as the stream holder. The DO refactor turned out to be heavier than expected: WebSocket objects cannot be serialized across DO `fetch()` boundaries, so `awaitTxid` would require a full WS upgrade at the DO level — a different protocol stack entirely. PHASES.md explicitly documents this as the acceptable fallback: "An acceptable fallback is a module-level singleton keyed by sender address." Placed on `SettlementService` (not truly module-level) for better lifecycle hygiene.

## Tests

- `src/__tests__/hiro-tx-stream.test.ts` — 11 tests (+7 new)
  - One WS for N concurrent subscribers on same instance
  - LRU eviction resolves oldest with null (polling fallback)
  - Factory rejection resolves subscribe() with null immediately (fixed synchronous drain race)
  - Resolver cleanup after terminal state (no double-resolve)
  - Reuse of open WS for post-open subscribers
  - close() drains all pending resolvers
  - **Burst: 20 concurrent subscribers → 1 WS**

- `src/__tests__/nonce-do-tx-stream.test.ts` — 4 new tests (settlement-level pooling)
  - 20 concurrent awaitTxid on same sender → 1 WS (core acceptance criterion)
  - Different senders each get their own WS (no cross-sender mixing)
  - Repeated calls reuse the same stream instance
  - WS factory rejection falls back gracefully for all subscribers

- Fixed pre-existing failure in `payment-validation.test.ts`: the assertion used `maxPollTimeMs=5000` which left 0ms for polling after stream budget. Corrected to `maxPollTimeMs=60_000`.

## Independence

This PR is independent of the Phase 1/2/3/4/5 attribution chain (PRs #381-#384, landing-page #883). It can merge in any order.

Closes #376.

### DIFF
--- a/src/__tests__/hiro-tx-stream.test.ts
+++ b/src/__tests__/hiro-tx-stream.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, vi } from "vitest";
 import type { Logger } from "../types";
 import {
   waitForHiroTxConfirmationViaStream,
+  HiroTxStream,
   type WebSocketFactory,
   type WebSocketLike,
 } from "../services/hiro-tx-stream";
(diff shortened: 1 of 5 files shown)


arc0btc (approved):
Pools Hiro WebSocket connections per sender address — one WS instead of one-per-call during bursts. The 156 WS 429 fallbacks over 4 days made this change necessary; the before/after numbers in the description confirm it works.

**What works well:**
- `HiroTxStream` class design is clean: lazy open, LRU cap at 256, factory injection for testability, drain-on-close safety net
- `SettlementService.getOrCreateStream()` is the right abstraction boundary — pool lives with the service, not the call
- Factory rejection → immediate null fallback is correct and tested
- `waitForHiroTxConfirmationViaStream()` kept as a thin one-shot wrapper preserves backward compat without any caller changes
- Test coverage is thorough — mock factory call counting is a clean way to assert the core invariant

**[suggestion] `sendQueue` appears to be dead code** (`src/services/hiro-tx-stream.ts`)

The `sendQueue: string[]` field is initialized, flushed in `handleOpen`, and cleared in `close()` — but nothing ever pushes to it. The loop in `handleOpen` that iterates `this.pending.keys()` covers the pre-open subscriber case already, so `sendQueue` is never populated. If this was a carry-over from an earlier design, it's safe to remove. If it was intentional as a message buffer for a different flow, a comment explaining why would help.

**[suggestion] Pooled timeout doesn't clean up the pending entry** (`src/services/settlement.ts:705-720`)

When the stream budget times out in the pooled path, `timeoutPromise` resolves null but the txid stays in `stream.pending` until one of: the WS closes, LRU evicts it, or a late `tx_update` arrives. In practice the LRU cap at 256 keeps this bounded, and late-arriving `tx_update` events will call `resolveOne()` on an already-GC'd promise chain (safe — JS ignores redundant resolves). Still, with high-volume senders under stream budget pressure, pending can fill with zombie entries and trigger LRU eviction for legitimate concurrent calls.

A targeted cleanup in the timeout handler would make the lifecycle explicit:

```typescript
const timeoutPromise = new Promise<null>((resolve) => {
  setTimeout(() => {
    this.logger.info("Pooled Hiro tx stream timed out; falling back to polling", {
      txid,
      senderAddress,
      streamBudgetMs,
    });
    // Remove this txid from the pool's pending map so the slot is released.
    stream.cancel(txid);
    resolve(null);
  }, streamBudgetMs);
});
```

This would require a `cancel(txid)` method on `HiroTxStream` that resolves and removes a single pending entry — a small addition but makes the lifecycle explicit.

**[nit] `FakeWebSocket` duplicated in test files**

`nonce-do-tx-stream.test.ts` and `hiro-tx-stream.test.ts` both define identical `FakeWebSocket` implementations. Could share a `src/__tests__/helpers/fake-websocket.ts` fixture.

**Code quality notes:**
- The `SettlementStreamPool` helper in `nonce-do-tx-stream.test.ts` mirrors `SettlementService.getOrCreateStream()` exactly by design — keeps the test self-contained and the assertion tight. Accepted.
- `buildHiroTxStreamUrl` export is minimal. Could be `getHiroTxStreamUrl` renamed, but the new name is clearer about intent from callers.

**Operational context:**
We track this relay's 429s in our dispatch logs. 156 fallbacks over 4 days translates to degraded settlement latency on about ~5% of bursts. This fix directly addresses the pattern. The Workers isolate lifecycle assumption (never-evict pool self-clears on eviction) is correct — we've seen the same pattern work well in our own sensor pools.

secret-mars (commented):
Substantive review — performance refactor (per-sender WS pooling for Hiro tx-update stream, closes #376). Already arc-APPROVED + codex/copilot-COMMENTED. The structural pivot from per-call WebSocket to per-sender pooled WebSocket is the right shape and the impact numbers (20 WS → 1 WS for 20 concurrent /settle on same sender) match the May 14 burst evidence. This review surfaces two resource-lifecycle issues that may bite under sustained traffic.

## What I checked

- `src/services/hiro-tx-stream.ts:90-322` — new `HiroTxStream` class (subscribe/close/drainPending/resolveOne, LRU eviction at maxPending=256)
- `src/services/hiro-tx-stream.ts:425-475` — backward-compat `waitForHiroTxConfirmationViaStream` wrapper (one-shot, maxPending=1, unconditional close after race)
- `src/services/settlement.ts:108-138` — `txStreams` pool + `getOrCreateStream`
- `src/services/settlement.ts:705-732` — pooled vs one-shot path in `awaitConfirmationPublic`
- `src/__tests__/hiro-tx-stream.test.ts` + `nonce-do-tx-stream.test.ts` — 403 lines of new tests covering multi-subscriber fan-out, eviction, error/close drain

## Findings

### 1. Pooled path leaks subscriber on timeout — orphaned `pending` entries accumulate until LRU eviction

**Where:** `src/services/settlement.ts:711-722`:

```ts
if (senderAddress) {
  const stream = this.getOrCreateStream(senderAddress);
  const subscribePromise = stream.subscribe(txid);
  const timeoutPromise = new Promise<null>((resolve) => {
    setTimeout(() => { ...; resolve(null); }, streamBudgetMs);
  });
  streamResult = await Promise.race([subscribePromise, timeoutPromise]);
}
```

When the timeout wins the race, `subscribePromise` is not cancelled and no cleanup is performed on the pooled stream. The resolver remains in `stream.pending[txid]` (and `txid` remains in `stream.subscribed`). If `tx_update` for that txid never arrives (the expected case when we've already fallen back to REST polling), the entry stays in `pending` until either:

- (a) The LRU eviction fires at `maxPending=256`, OR
- (b) A late `tx_update` arrives and `resolveOne(txid, result)` clears it

Compare the one-shot path at `hiro-tx-stream.ts:438-461` — it calls `stream.close()` after the race regardless of winner, which calls `drainPending(null)` and clears everything. The pooled path has no analogous cleanup.

**Consequences:**
- Under steady-state burst-then-quiet, `pending` fills with ghost entries. Worst case: 256 ghosts before LRU kicks in.
- Each LRU eviction emits `"HiroTxStream LRU eviction"` WARN. The new tests don't check eviction-cause provenance; ops alerting on this WARN will see false positives from ghost entries that were never real over-capacity.
- Memory cost is small (a Map entry + a single resolver fn ref) but unbounded *over the SettlementService lifetime* per sender stream.

**Suggested fix:** Expose an `unsubscribe(txid, resolver)` method on `HiroTxStream` that removes the specific resolver from the pending list (and removes the entry entirely + sends an unsubscribe-from-Hiro if no resolvers remain). Call it on timeout in the pooled path:

```ts
const subscriber = stream.subscribe(txid);  // capture the resolver-bound promise
const timeoutId = setTimeout(...);
const result = await Promise.race([subscriber, timeoutPromise]);
clearTimeout(timeoutId);
stream.unsubscribe(txid /* and the specific resolver if accumulating */);
```

Roughly +20 LOC including method + call-site cleanup. Test would assert `stream.pendingCount === 0` after a timeout race.

### 2. `txStreams` Map has no eviction — WSes held indefinitely per sender

**Where:** `src/services/settlement.ts:112-119` (comment + initialization):

> *"Entries are created lazily and never evicted — Cloudflare Workers isolates are short-lived, so in-memory state self-clears on eviction without leaking sockets beyond the Worker's lifetime."*

This reasoning holds for short-lived Worker invocations on the request path. But the settlement code also runs inside Durable Objects (the `SchedulerDO` alarm path you've been wiring tenero into per [lp#794](https://github.com/aibtcdev/landing-page/issues/794)/[lp#880](https://github.com/aibtcdev/landing-page/issues/880)), which can run for minutes and accumulate state across alarm fires. If a long-lived DO instance receives `/settle` traffic for N distinct senders before being evicted, it holds N open WebSockets to Hiro indefinitely — even for senders with no traffic for hours.

Pre-pool design closed the WS at end of each settle (per-call open/close). New design *only* closes via explicit `stream.close()`, which the pooled path never calls.

**Consequences:**
- Hiro free-tier connection limits per IP may be silently hit as the pool grows
- Idle WSes are still being kept alive by browser-style ping/pong (depending on Hiro's server); if Hiro drops idle connections, the next subscribe will trigger reconnect anyway — so idle WS holding is wasted work
- The `txStreams` Map itself has no bound — memory grows linearly with distinct senders seen

**Suggested fix:** Add an idle-TTL eviction (e.g. close + remove if no subscribe activity in last N minutes), or a hard cap on Map size with LRU eviction calling `stream.close()` on the evicted entry. Easiest: piggyback on each `getOrCreateStream` call — sweep entries whose `pendingCount === 0 && lastUsedAt < now - IDLE_TTL_MS` and close them. ~15 LOC.

If the design intent is "trust isolate lifetime as the eviction signal," call that out explicitly and acknowledge the DO case — perhaps as a constraint that DO-borne SettlementService instances should call `closeAllStreams()` on DO hibernation alarm.

## Smaller notes (non-blocking)

- **`handleError` doesn't call `socket.close()`** — only nulls `this.socket` and clears state. The underlying WS may still be open. Usually the runtime will fire close after error so this self-resolves, but explicit `this.socket?.close()` before nulling would be tidier and avoid the small window where a second `openSocket()` creates a duplicate before the first closes.

- **Subscription dedup is per-txid, not per-(sender,txid)** — `HiroTxStream` is already scoped per-sender so this works, but a comment on `subscribe()` saying "txid uniqueness assumed within sender scope" would help anyone reading just the class.

## Verdict

Structural shape is correct and the impact numbers earn the change. Finding #1 (timeout-leak) is the more concretely-observable concern — easy to test and adds bounded protection against ghost-entry accumulation. Finding #2 is a longer-tail risk that matters once DO-borne SettlementService instances become common. Both are addable in follow-up PRs without blocking this merge if your preference is to land the perf win now and tighten lifecycle in a sequel.

Not approving this round — the pooled-path leak is a behavior the new design introduces that the per-call design didn't have, and it's worth a small fix before merge OR a tracking issue with a clear owner. Happy to re-review once either lands.

PR #383, opened 2026-05-19
feat(sponsor): add sponsorNonceValidForMs and docs to nonce-expires-at contract (#374)

whoabuddy:
## Summary

Stacks on top of PR #379 (via PR #380) to complete the two-field wire contract for
the sponsor nonce TTL defined in PHASES.md for issue #374.

**Relationship to existing PRs:**
- **PR #379** (`feat/374-sponsor-nonce-expires-at`): added `nonceExpiresAt` to `/relay` responses and threaded it through NonceDO, SponsorService, and okWithTx. This PR depends on that work.
- **PR #380** (`fix/380-fallback-nonce-expiry-constant`): added `FALLBACK_NONCE_EXPIRY_MS` constant and a base-endpoint test. This PR is based on that branch.
- **This PR** (`feat/374-sponsor-nonce-ttl-contract`): adds `sponsorNonceValidForMs`, surfaces `nonceExpiresAt` on `/sponsor` explicitly (PR #379 only threaded it to `/relay`), adds docs, and updates OpenAPI + discovery.

**Field naming decision:**
- Kept `nonceExpiresAt` from PR #379 (already in review; renaming would break that PR).
- Added `sponsorNonceValidForMs` as the second field per PHASES.md — equals `STALE_THRESHOLD_MS` = 600 000 ms.
- Deviation from PHASES.md: PHASES.md uses `sponsorNonceExpiresAt`; this PR uses `nonceExpiresAt` (from #379). Both fields carry the same semantic; the naming follows the established PR #379 contract to avoid churn.

## Changes

| File | What changed |
|------|--------------|
| `src/endpoints/sponsor.ts` | `SPONSOR_NONCE_VALID_FOR_MS` constant; both fields in `/sponsor` 200 response and Chanfana schema |
| `src/services/sponsor.ts` | Fix `fetchNonceFromDO` return type to include `nonceExpiresAt` in success branch (TS error from #379's wire-up) |
| `docs/sponsor-nonce-ttl.md` | **NEW** — contract doc: fields, source constant, consumer adoption checklist, clock-skew guidance |
| `src/routes/discovery.ts` | `/llms-full.txt`: update `/sponsor` success example + add TTL subsection. `topics/sponsored-transactions`: add both fields to `/relay` and `/sponsor` examples + add "Sponsor Nonce TTL" section with retry-queue pattern |
[...]

Issue #377, opened 2026-05-15
Use stacks-core reason_data.is_origin for nonce-conflict attribution in settlement.ts and recovery paths


whoabuddy:
## Summary

The relay has a typed attribution resolver at `src/utils/broadcast-outcome.ts:28-136` that correctly maps stacks-core's `reason_data.is_origin` to `responsible: \"sender\" | \"sponsor\"` for `ConflictingNonceInMempool` and other rejection reasons. **The hot path in `src/services/settlement.ts` does not use it.** Instead, settlement.ts uses `sponsorNonceForLog === null` as a proxy for \"this is the client's problem,\" which conflates two unrelated things:

- \"Did *this endpoint* assign a sponsor nonce on *this call*?\"
- \"Was the conflict on the sender side?\"

For pre-sponsored txs arriving at `/settle` (the path described in #373), `sponsorNonceForLog` is `null` but the conflict can still be sponsor-side. We currently log these as INFO \"client pre-signed tx\" when they're actually sponsor-side bugs.

We control the sponsor wallet; we must accept the client's submitted tx as truth and use the node's response to attribute. The only authoritative signal is `reason_data.is_origin`.

## Current behavior

`src/services/settlement.ts:858-897`:

```ts
if (isNonceConflict) {
  ...
  if (sponsorNonceForLog === null) {
    // No relay-assigned sponsor nonce: this is a client pre-signed tx.
    // Log at INFO — the nonce conflict is the client's problem, not the relay's.
    this.logger.info(\"Broadcast rejected due to client nonce conflict (pre-signed tx)\", { ... });
  } else {
    // Relay assigned a sponsor nonce: unexpected conflict on relay side.
    this.logger.warn(\"Broadcast rejected due to nonce conflict\", { ... });
  }
  return { ..., nonceConflict: true, clientRejection: matchedReason, ... };
}
```

Issues with this:
1. The branch decision uses the wrong signal (`sponsorNonceForLog === null` is about *bookkeeping in this call*, not about *who caused the conflict*).
2. `is_origin` is parsed into `reasonData` already (line 836) but never consulted for `ConflictingNonceInMempool`. It IS consulted correctly for `TooMuchChaining` at line 904.
3. `BadNonce` is in `NONCE_CONFLICT_REASONS` (`src/utils/stacks.ts:10`) and falls into the same branch with no `is_origin` check, even though stacks-core provides it.

Downstream consumers can't distinguish sender-fault from sponsor-fault:
[...]

PR #386, opened 2026-05-19
docs(quest): verification prep for nonce-conflict-attribution

whoabuddy:
## Summary

Pre-deploy verification checklist and log query runbook for the nonce-conflict-attribution quest (issues #373–#377).

- Baseline metrics captured from `logs.aibtc.com` as of 2026-05-18 (pre-deploy)
- Exact `curl` log queries for each acceptance criterion against `logs.aibtc.com`
- Merge-order guidance for the 4-deep PR chain (#379→#380→#383→#384) and independent PRs (#381, #382, #385)
- Note on landing-page PR #883 timing (must merge after relay #383 is live on mainnet)
- Issue closure comment drafts for #373, #374, #375, #376, #377 (with TODO placeholders for observed metrics)

This file will be updated in-place once all PRs merge and the 24h production window closes.

## Related PRs (all OPEN)

Relay: #379, #380, #381, #382, #383, #384, #385
Landing-page: aibtcdev/landing-page#883

## Test plan

- [ ] Review merge-order guidance in §2
- [ ] Confirm worker-logs sanity check in §3 is reproducible
- [ ] Verify baseline metrics in §4 match production reality
- [ ] After all PRs merge: run log queries from §5 at T+24h
- [ ] Fill in TODO placeholders in §8 issue closure drafts
- [ ] Close issues #373–#377 with the draft comments

### DIFF
--- a/docs/nonce-conflict-attribution-verification.md
+++ b/docs/nonce-conflict-attribution-verification.md
@@ -1,0 +1,394 @@
+# Verification: nonce-conflict-attribution
+
+**Prepared:** 2026-05-18
+**Status:** Pre-deploy — PRs open, 24h window not yet started
+**Quest:** nonce-conflict-attribution
+
+---
+
+## 1. Pre-Deploy State
+
+### Production relay version at time of prep
+
+```
+GET https://x402-relay.aibtc.com/health → version: "1.32.1"
+GET https://x402-relay.aibtc.dev/health  → version: "1.32.1" (staging)
+```
+
+None of the quest PRs are deployed yet. All seven PRs are open as of 2026-05-18.
+
+### PR Dependency Stack
+
+| PR | Title | Repo | Status | Base branch | Issue |
+|----|-------|------|--------|-------------|-------|
+| [#379](https://github.com/aibtcdev/x402-sponsor-relay/pull/379) | feat(relay): add nonceExpiresAt to /relay and /sponsor responses (#374) | relay | OPEN | main | #374 |
+| [#380](https://github.com/aibtcdev/x402-sponsor-relay/pull/380) | fix(sponsor): FALLBACK_NONCE_EXPIRY_MS constant + okWithTx nonceExpiresAt test | relay | OPEN | feat/374-sponsor-nonce-expires-at (#379) | #374 |
+| [#383](https://github.com/aibtcdev/x402-sponsor-relay/pull/383) | feat(sponsor): add sponsorNonceValidForMs and docs to nonce-expires-at contract (#374) | relay | OPEN | fix/380-fallback-nonce-expiry-constant (#380) | #374 |
+| [#384](https://github.com/aibtcdev/x402-sponsor-relay/pull/384) | test(sponsor): contract test for nonceExpiresAt TTL (#375) | relay | OPEN | feat/374-sponsor-nonce-ttl-contract (#383) | #375 |
+| [#381](https://github.com/aibtcdev/x402-sponsor-relay/pull/381) | fix(settlement): attribute nonce conflicts via reason_data.is_origin (#377) | relay | OPEN | main | #377 |
+| [#382](https://github.com/aibtcdev/x402-sponsor-relay/pull/382) | fix(settle): re-sponsor pre-sponsored txs on sponsor-fault conflict (#373) | relay | OPEN | main | #373 |
+| [#385](https://github.com/aibtcdev/x402-sponsor-relay/pull/385) | perf(settle): pool Hiro WebSocket subscriptions per sponsor sender (#376) | relay | OPEN | main | #376 |
+| [#883](https://github.com/aibtcdev/landing-page/pull/883) | fix(inbox): clamp reconciliation queue to nonceExpiresAt (#375) | landing-page | OPEN | main | #375 |
+
+### PR Chain Dependency Diagram
+
+```
+main
+ ├── #379 (base: main) ─► #380 (base: #379) ─► #383 (base: #380) ─► #384 (base: #383)
+ │                                              [Phase 3 / #374]     [Phase 4 / #375]
+ ├── #381 (base: main)   [Phase 1 / #377]
+ ├── #382 (base: main)   [Phase 2 / #373]  ← logically depends on #381 being deployed first
+ └── #385 (base: main)   [Phase 6 / #376]
+
+landing-page:
+ └── #883 (base: main)   [Phase 5 / #375]  ← requires relay #383 deployed to prod
+```
+
+**Note on #381 and #382:** Both have `main` as base branch. Phase 2 (#382) was developed assuming Phase 1 (#381) code is present. Merge #381 first, then #382 should be rebased/verified before merge.
+
+---
+
+## 2. Merge Order Guidance
+
+### Recommended sequence
+
+1. **#379** — Merge first. Foundation for the #374 TTL chain. `feat/374-sponsor-nonce-expires-at` → `main`
+2. **#380** — Merge immediately after #379. Stacked on #379; will need rebase after #379 merges. `fix/380-fallback-nonce-expiry-constant` → `main`
+3. **#383** — Merge after #380. Stacked on #380; rebase after #380 merges. `feat/374-sponsor-nonce-ttl-contract` → `main`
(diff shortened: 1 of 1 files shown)


arc0btc (approved):
Pre-deploy verification runbook for the nonce-conflict-attribution quest — baseline metrics captured, exact log queries ready, merge order documented.

**What works well:**
- §4 baseline metrics are excellent: specific event counts with exact time windows give post-deploy comparison a clean anchor. Knowing the `responsible` field is null on all 10 pre-deploy events (§10 note 1) prevents false negatives when querying post-deploy.
- The merge-order diagram in §2 is exactly right for a 4-deep stacked PR chain. The call-out that #382 has `main` as its base but logically depends on #381 being deployed first is the kind of detail that prevents a painful mid-deploy revert.
- §10 notes are high signal — the WS pooling note (module-level singleton rejected DO approach due to non-serializable WebSocket) documents a non-obvious constraint that would otherwise look like a bug later.
- Issue closure drafts in §8 follow a good pattern: fix reference + observable metric + TODO placeholder for post-deploy fill-in.

**[suggestion] Log query `limit=500` could silently miss data** (§5, all metrics)
All seven queries cap at 500 results. For low-count metrics (CONFLICTING_NONCE, WS 429 fallbacks) this is safe. But Metric 2's second query scans all DEBUG-level x402-relay events for nonce assignment text — a busy 24h window could have >500 DEBUG events before any nonce assignment logs appear. Worth a note about this constraint and how to paginate (`offset` parameter if the logs API supports it, or narrowing the time window) so whoever runs this post-deploy doesn't get a silent undercount.

**[question] Metric 2 cross-check assumes sponsor nonce log vocabulary**
The second jq filter in Metric 2 relies on `.message | contains("nonce") and contains("assign")` at DEBUG level. Is that log message pattern confirmed from the source? If the phrasing in #379/#383 differs (e.g. "nonce.assigned" or "sponsorNonce: reserved"), this query returns 0 and the cross-check vacates silently. Worth pinning the exact message string from the diff before post-deploy verification.

**[nit]** §3 sanity check uses `~/dev/aibtcdev/worker-logs/.env` — this is whoabuddy's local path. Not wrong for a personal runbook, but a parenthetical "(adapt to your local worker-logs checkout)" would make it usable by anyone else on the team without silent confusion.

**Code quality notes:**
Docs-only change. The jq is well-formed throughout — `group_by(.timestamp[0:16])` for minute-bucketing in Metric 5 is clean. The acceptance table in §7 maps each metric to its query section, which is the right structure for a runbook that will be referenced under post-deploy time pressure.

*Operational note:* We run x402 payment flows operationally and have seen the WS 429 fallback rate spike during settlement bursts. The module-level singleton approach in #385 (LRU-capped at 256 per sender) should collapse the per-burst count significantly — baseline of 7 in a single minute is consistent with unbounded WS opens on concurrent settles.

secret-mars:
Cross-linking my substantive reviews on the related PRs since this doc is the integration point for the whole quest. Each affects specific verification metrics in §5.

## My reviews on quest PRs

| PR | My review | State | Load-bearing finding |
|---|---|---|---|
| [#381](https://github.com/aibtcdev/x402-sponsor-relay/pull/381#pullrequestreview-4315000166) | substantive | APPROVED | LP `lib/inbox/relay-rpc.ts` parsers don't extract `responsible` + `agentErrorCode` — fields cross relay→LP RPC boundary but get dropped |
| [#382](https://github.com/aibtcdev/x402-sponsor-relay/pull/382#pullrequestreview-4314857445) | substantive | COMMENTED | Stats double-fire in recovery branches (parent path + sub-path both `logFailure`); base-mismatch bundles ~290 LOC of #381 |
| [#385](https://github.com/aibtcdev/x402-sponsor-relay/pull/385#pullrequestreview-4315051806) | substantive | COMMENTED (non-approve) | Pooled `awaitConfirmationPublic` path doesn't unsubscribe on timeout → orphans accumulate until LRU eviction at maxPending=256; `txStreams` Map has no eviction |
| [lp#883](https://github.com/aibtcdev/landing-page/pull/883#pullrequestreview-4314947260) | substantive | COMMENTED | `parseSubmitPaymentResult` doesn't extract `nonceExpiresAt` → staging uses LP local clock, not relay's authoritative timestamp; `isSponsorNonceExpired`'s "relay clock authoritative" branch is dead code |

## Where these intersect §5 verification metrics

### Metric 1 (ConflictingNonceInMempool where responsible=sponsor)
The query filters on `.context.responsible == "sponsor"`. Confirm whether the WARN log line on `relay#381` (`settlement.ts` attribution wiring) writes `responsible` to `.context.responsible` exactly — the field is added on `BroadcastOnlyResult` but the log call in `settle.ts:566-568` (parent path) doesn't include it; only the new INFO `"Broadcast failure attribution"` log at `settle.ts:560` does. If §5.1's filter expects the older WARN line to carry the field, queries will return 0 even when sponsor-side conflicts happen.

### Metric 2 (SENDER_NONCE_CONFLICT events; no sponsor slot burned)
The second cross-check (`.message contains "nonce" and "assign"`) works against relay logs but the first query also expects `.context.agentErrorCode == "SENDER_NONCE_CONFLICT"`. The relay sets this on the broadcast result, but the LP-side `lib/inbox/relay-rpc.ts` doesn't extract `agentErrorCode` (my #381 finding) — so `aibtc-landing` logs won't carry it, only `x402-relay` will. Worth adding a note that this metric is relay-side only, or scoping the LP-side query to `errorReason == "conflicting_nonce"` granularly.

### Metric 4 (/sponsor response includes nonceExpiresAt)
Direct probe via `test:sponsor` or OpenAPI spec confirms the relay surfaces it. But the LP-side consumption probe is worth adding: `grep` `lib/inbox/relay-rpc.ts` for `nonceExpiresAt` references — currently zero (my #883 finding). Even if /sponsor returns the field, the LP won't consume it until that parser extension lands. This means Metric 7's "re-sponsor calls past TTL" trigger is based on LP-derived `stagedAt + SPONSOR_NONCE_TTL_MS`, not the relay's authoritative `nonceExpiresAt`.

### Metric 5 (WS 429 fallbacks ≤1 per 20-tx burst)
The query greps `Hiro tx stream errored; falling back`. After #385 merges, the per-call WARN log message changes to `HiroTxStream: WebSocket error; draining pending` and the LRU-eviction emits `HiroTxStream LRU eviction; falling back to polling`. Two consequences:
1. The current query message-string won't match post-deploy — needs update to either both strings or the new shape
2. If my #385 pooled-path leak finding bites, ghost-entry LRU evictions will inflate the count without representing real over-capacity

### Metric 6 (No new SENDER_NONCE_GAP)
Looks correct as written — wouldn't be affected by the wire-extraction gaps since SENDER_NONCE_GAP is pre-existing.

### Metric 7 (LP re-sponsor calls past TTL)
The `grep` for `re-sponsor | sponsor_refresh | nonce_expired` matches the actual log messages from lp#883 (`resubmit_after_nonce_expiry`, `resubmit_succeeded`, `ack_nonce_expired_no_tx_hex`). One pattern missing: `ack_resubmit_kv_write_failed` — the orphan case I flagged in #883 finding #4. Worth adding to the grep so any non-zero counts surface as ops alerts.

## Suggested additions to §8 (issue closure drafts)

For **#374 closure**, add a sentence noting that LP-side `parseSubmitPaymentResult` extension is a separate follow-on (Phase 5.1) so the wire contract is end-to-end consumed.

For **#377 closure**, same point about LP-side `responsible` + `agentErrorCode` extraction being deferred — otherwise the attribution data is captured but never reaches consumer observability.

For **#376 closure** (Metric 5), confirm the pooled-path subscription leak from my #385 finding #1 is addressed (or tracked) before closing — otherwise the "long-tail accumulation" risk remains.

## Doc-level observations

- §3 sources `~/dev/aibtcdev/worker-logs/.env` for admin key — local-machine assumption that won't help anyone reading the doc later (e.g. a future maintainer or me). Worth either inlining how to retrieve the key OR pointing to where in 1Password / secrets store it lives
- §1 "PR Dependency Stack" table notes #381/#382/#385 base on `main` independently — matches what I found reviewing #382 (my finding #1 there)

Happy to open the LP-side parser extension as Phase 5.1 issues against `aibtcdev/landing-page` once #381 + #883 merge, if no one else picks them up.

PR #390, opened 2026-05-19
fix(settle): gate buggy re-sponsor recovery behind ENABLE_SETTLE_RESPONSOR (default off)

whoabuddy:
## Summary

Production incident **2026-05-19T13:00 UTC** stuck 3 wallet-0 sponsor nonces (3854, 3855, 3856) for 22+ min with `"Cannot sponsor sign a non-sponsored transaction"` retries. This is the second occurrence of a regression introduced by #382 (Phase 2 of nonce-conflict-attribution quest, issue #373).

This PR **gates the buggy re-sponsor branch behind `ENABLE_SETTLE_RESPONSOR` (default `"false"`)**, restoring pre-Phase-2 behavior: sponsor-fault conflicts on pre-sponsored `/settle` calls return `CONFLICTING_NONCE` (retryable, no slot burned) instead of attempting recovery.

## Root cause

The re-sponsor branch at `src/endpoints/settle.ts` calls `SponsorService.sponsorTransaction(parsedTx)` where `parsedTx` is the original *already-sponsored* hex. That routes through the hand-submit dispatch queue (`boundedBroadcast`), which then attempts to sponsor-sign a tx that the worker treats as non-sponsored — wedging the sponsor nonce slot until `nonce_reconcile_stale` clears it ~22 min later.

| | Burst 1 | Burst 2 |
|---|---|---|
| Time | 2026-05-19T07:00 UTC | 2026-05-19T13:00 UTC |
| Stuck nonces | 1 (wallet 1 / 1811) | 3 (wallet 0 / 3854-3856) |
| Duration | 22 min, self-healed | 17+ min (cleared by manual flush-wallet) |
| Severity | bounded | escalating |

## What this PR does NOT change

- The other quest deliverables (#377 attribution wiring, #374 TTL contract, #376 WS pool) remain fully active — they're not implicated.
- The sender-fault path (returns `SENDER_NONCE_CONFLICT`) is unaffected.
- The auto-sponsor recovery path (sponsorNonce !== null) is unaffected.
- The re-sponsor code itself is preserved — only the entry gate is added. A proper fix can re-enable it.

## Why a flag rather than full revert

The flag preserves the test infrastructure, the `SENDER_NONCE_CONFLICT` error code, the stats categorization, and the surrounding logging. A follow-up PR can rewrite the re-sponsor implementation to bypass the dispatch queue + clean up its own queue entry on failure, then re-enable the flag — without re-doing the scaffolding.

## Production cleanup performed before this PR

- Manual `POST /nonce/reset {action:"flush-wallet", walletIndex:0}` cleared the 3 wedged nonces on wallet 0.
- New admin API key minted (previous one had expired); see notes in REGRESSION-NOTES.md in `.planning/2026-05-18-nonce-conflict-attribution/`.

## Changes

| File | Change |
|------|--------|
| `src/endpoints/settle.ts` | Gate the responsor branch on `c.env.ENABLE_SETTLE_RESPONSOR === "true"`. When off, sponsor-fault falls through to CONFLICTING_NONCE with a clear log line. |
| `src/types.ts` | Declare optional `ENABLE_SETTLE_RESPONSOR` env var with full doc explaining why it defaults off. |
| `src/__tests__/settle-responsor.test.ts` | 4 new tests for the flag gate. |
| `wrangler.jsonc` | Declare `"ENABLE_SETTLE_RESPONSOR": "false"` in dev / staging / production env blocks. |

## Test plan

- [x] `npm run check` clean
- [x] `npm test -- settle-responsor` — all 15 tests pass (11 existing + 4 new)
- [x] Full `npm test` — only pre-existing `nonce-do-sponsor-status` failure on main (unrelated)
- [ ] CI green on this PR
- [ ] Post-deploy: 0 new `Re-sponsor failed` events
- [ ] Post-deploy: 0 new `"Cannot sponsor sign a non-sponsored transaction"` errors

## Follow-up

File a tracking issue for the proper re-sponsor rewrite. Rough approach:

1. Mint a sponsor nonce manually (without hand-submit).
2. Sign with `@stacks/transactions sponsorTransaction()` in-process.
3. Broadcast directly via `settlementService.broadcastOnly()`.
4. On any failure: release the nonce, do NOT enqueue, return CONFLICTING_NONCE.
5. Add a vitest case that mocks the sign step to throw and asserts no dispatch_queue side-effect.
6. Re-enable the flag.

Refs #373, #382. cc @arc0btc

### DIFF
--- a/src/__tests__/settle-responsor.test.ts
+++ b/src/__tests__/settle-responsor.test.ts
@@ -205,3 +205,50 @@ describe("Non-conflict broadcast failure does not enter recovery branches", () =
     expect(shouldEnterPreSponsoredPath).toBe(false);
   });
 });
+
+// ---------------------------------------------------------------------------
+// 6. ENABLE_SETTLE_RESPONSOR feature flag — gates the buggy re-sponsor path
+// ---------------------------------------------------------------------------
+
+describe("ENABLE_SETTLE_RESPONSOR feature flag", () => {
+  function shouldEnterResponsorBranch(
+    broadcastResult: BroadcastOnlyResult,
+    sponsorNonce: number | null,
+    envFlag: string | undefined
+  ): boolean {
+    const enabled = envFlag === "true";
+    const inPreSponsoredGate =
+      sponsorNonce === null &&
+      ((broadcastResult as { nonceConflict?: boolean }).nonceConflict === true ||
+        (broadcastResult as { tooMuchChaining?: boolean }).tooMuchChaining === true);
+    return (
+      inPreSponsoredGate &&
+      (broadcastResult as { responsible?: "sender" | "sponsor" | "network" }).responsible === "sponsor" &&
+      enabled
+    );
+  }
+
+  it("OFF by default: sponsor-fault conflict on pre-sponsored tx does NOT enter re-sponsor branch", () => {
+    const broadcastResult = makeBroadcastOnlyError("sponsor");
+    expect(shouldEnterResponsorBranch(broadcastResult, null, undefined)).toBe(false);
+    expect(shouldEnterResponsorBranch(broadcastResult, null, "false")).toBe(false);
+    expect(shouldEnterResponsorBranch(broadcastResult, null, "")).toBe(false);
+  });
+
+  it("ON (\"true\"): sponsor-fault conflict on pre-sponsored tx enters re-sponsor branch", () => {
+    const broadcastResult = makeBroadcastOnlyError("sponsor");
+    expect(shouldEnterResponsorBranch(broadcastResult, null, "true")).toBe(true);
+  });
+
+  it("Flag does NOT affect sender-fault path (always returns SENDER_NONCE_CONFLICT)", () => {
+    const broadcastResult = makeBroadcastOnlyError("sender", "sender_nonce_confirmed");
+    expect(shouldEnterResponsorBranch(broadcastResult, null, "true")).toBe(false);
+    expect(shouldEnterResponsorBranch(broadcastResult, null, "false")).toBe(false);
+  });
+
+  it("Flag does NOT affect auto-sponsored path (sponsorNonce !== null)", () => {
+    const broadcastResult = makeBroadcastOnlyError("sponsor");
+    expect(shouldEnterResponsorBranch(broadcastResult, 42, "true")).toBe(false);
+    expect(shouldEnterResponsorBranch(broadcastResult, 42, "false")).toBe(false);
+  });
+});
--- a/src/endpoints/settle.ts
+++ b/src/endpoints/settle.ts
@@ -672,7 +672,18 @@ export class Settle extends BaseEndpoint {
         ) {
           // Pre-sponsored tx (sponsorNonce === null) had a nonce conflict.
           // Use Phase 1's responsible signal to determine who caused the conflict.
-          if (broadcastResult.responsible === "sponsor") {
+          //
+          // The sponsor-fault re-sponsor path is gated behind ENABLE_SETTLE_RESPONSOR
+          // because the current implementation routes the pre-sponsored tx through
+          // the hand-submit dispatch queue, which then attempts to sponsor-sign a
+          // tx that the broadcast worker treats as non-sponsored — wedging the sponsor
+          // nonce slot until the next reconcile_stale (~22 min). Production incident
+          // 2026-05-19T13:00 UTC stuck 3 wallet-0 nonces. Until the path is rewritten
+          // to bypass the dispatch queue and clean up its own queue entry on failure,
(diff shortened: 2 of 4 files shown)


secret-mars:
Fast-fire ack on the rollback gate — this is the right shape given two incidents in 6h with escalating wedge depth (1→3 nonces) and external clearance now required. Validates that the v420 review-cycle concerns on #382's recovery path (re: stats double-fire) were directional but missed the load-bearing wedge — my apologies for the depth gap on that one. Adding two consumer-side observations in case they shape the follow-up rewrite.

## LP consumer-side: does the gated path leak through to reconciliation-queue?

After #883 merged, [`lib/inbox/reconciliation-queue.ts:261-313`](https://github.com/aibtcdev/landing-page/blob/main/lib/inbox/reconciliation-queue.ts#L261) is the resubmit path:

```ts
resubmitResult = await rpc.submitPayment(...);
// ...
if (!resubmitResult.accepted) {
  // Relay rejected the re-submission — ack and discard.
  logPaymentEvent(logger, "warn", "payment.retry_decision", ...);
  message.ack();
  continue;
}
```

With `ENABLE_SETTLE_RESPONSOR=false` (the new default), any sponsor-fault on a pre-sponsored `/settle` returns `CONFLICTING_NONCE`. Two unknowns from the LP-side view:

1. **Does `rpc.submitPayment` (JSON-RPC `/relay` binding) hit the same `src/endpoints/settle.ts` recovery code path the flag gates, or a different handler?** If the same path, every TTL-expired resubmit on a stuck pre-sponsored tx now returns rejected with `code: CONFLICTING_NONCE` to the LP — which currently triggers `ack-and-discard`. That would silently drop in-flight payments that *would* have been recoverable by mempool drain.
2. **If LP-side does need to handle `code: CONFLICTING_NONCE` differently from other rejection codes** (mempool-collision is retryable, not terminal), the resubmit branch's `ack-and-discard` becomes a consumer-side gap — the inline comment "treat submit failure as a terminal condition" was designed for the case where the relay was recovering internally.

Both unknowns are answerable by a 30-sec relay-side code check; I'm flagging them in case it's worth a quick LP-side defensive PR (gate the `ack-and-discard` on `code != CONFLICTING_NONCE`, requeue with backoff for that code class).

## Cross-link: lp#885 captures non-load-bearing #883 findings

[lp#885](https://github.com/aibtcdev/landing-page/issues/885) (filed today) tracks two unrelated post-merge edge cases from my pre-merge review on #883: retry cadence after fresh resubmit + ack-on-KV-write-failure observability. If the LP-side follow-up from this incident touches `reconciliation-queue.ts`, it might be worth folding finding A (retry cadence) into the same PR since both touch the resubmit branch.

## On the rewrite plan

The 6-step approach in the body looks sound — particularly **step 4 ("On any failure: release the nonce, do NOT enqueue, return CONFLICTING_NONCE")** which makes the failure mode self-clearing instead of wedging until reconcile. The vitest mock for sign-throw (step 5) is a good shape; could also assert no `boundedBroadcast` queue side-effect to lock in the regression-proof.

secret-mars:
Self-resolving the unknown from above after a quick relay-side + LP-side trace:

**Finding: LP path is NOT affected by the gate flip.** The gated re-sponsor branch is in `src/endpoints/settle.ts:676` — it fires when the `/settle` HTTP handler is called for a payment whose tx is already-sponsored (recipient calling /settle twice, or X-PAYMENT-replay scenarios). The LP reconciliation queue's resubmit goes through `rpc.submitPayment` ([`src/rpc.ts:105`](https://github.com/aibtcdev/x402-sponsor-relay/blob/main/src/rpc.ts#L105)) which routes through the PAYMENT_QUEUE → queue-consumer → `sponsorService.sponsorTransaction(...)` on a freshly-deserialized tx. LP stages `paymentPayload?.payload?.transaction` ([`route.ts:1543`](https://github.com/aibtcdev/landing-page/blob/main/app/api/inbox/%5Baddress%5D/route.ts#L1543)) which is the sender-signed-but-NOT-sponsored hex, so each rpc.submitPayment gets a fresh sponsor nonce assignment via the normal path. No re-sponsor of an already-sponsored tx ever happens on LP-reconciliation flow → no wedge surface.

So the consumer-side observations from my prior comment downgrade from "needs defensive PR" to "verified safe." Sorry for the over-cautious flag — the relay surface is segmented enough that the gate is correctly scoped to the /settle handler's recipient-retry branch.

**What remains worth tracking:** the rewrite plan in your PR body (steps 1-6) is a clean separation between sponsor-nonce assignment and broadcast — if you do file the tracking issue, happy to scout the LP consumer-side again once a re-enabled flag has been on staging for a non-trivial window. The signal-noise on LP's current resubmit path (lp#883's `ack-and-discard` on accepted=false) is fine as-is for the gate-OFF default.

Issue #398, opened 2026-05-23
Payments orphaned at "queued" with no on-chain tx → wallet wedged, message never delivered


biwasxyz:
## Summary

A payment can be permanently orphaned at `status: "queued"` with **no on-chain transaction and no live work item anywhere** — not in the mempool, not in a sender-hand, no in-flight queue message, no txid. The sender's nonce never advances, and because the aibtc.com inbox **dedups** repeat sends onto the existing pending payment, the agent's wallet becomes **wedged**: it can no longer send any message, even though it has sBTC and a healthy nonce.

This is distinct from #397. In #397 the tx **confirms on-chain** but the record never gets linked (a `txid_map` gap). Here the tx is **never broadcast at all** — the queue message is lost/exhausted and the record is left non-terminal. Both surface identically as `queued`.

## Live reproduction (mainnet, 2026-05-23)

Sent an inbox message from `SP4DXVEC…` ("Secret Mars", 2233 sats sBTC, nonce healthy) to `SP20GPDS…` via `send_inbox_message`. The API returned `success: true / "Message delivered" / "Payment accepted"`, but:

- **Relay** `/payment/pay_8890512cff5c415dbafa415f54c37948`: `status: queued`, `relayState: queued`, **no txid**, `submittedAt: 2026-05-23T06:37:13Z` (≈7h before the send — the inbox **deduped** onto the existing pending payment instead of creating a fresh one).
- **Chain** (`/address/SP4DXVEC…/nonces`): `last_executed: 864`, `possible_next: 865`, `detected_mempool_nonces: []` → **nonce 865 was never broadcast**.
- **Relay** `/nonce/state`: **no sender-hand entry** for `SP4DXVEC…` → not held, not pending anywhere.
- Full KV record: `sponsorWalletIndex: 0, sponsorNonce: 3869, broadcastingAt: 06:37:16, attempts: 1` → it went `broadcasting` exactly once, the broadcast failed/re-queued, and it **never advanced again**.

Net: the wallet is wedged at nonce 865. Every new message re-attaches to the dead `pay_8890512c`; the message will never deliver on its own.

## Root cause

The queue-consumer (`src/queue-consumer.ts`) has no guarantee that a payment reaches a **terminal** state. Two orphaning paths leave a record stuck at `queued`:

[...]

Issue #397, opened 2026-05-23
Payments stuck at "queued" on nonce-conflict → inbox messages paid but never delivered


biwasxyz:
## Summary

Payments that confirm on-chain via a **nonce-conflict** path are never linked back to their `payment:` record, so `/payment/:id` is stuck at `queued` (or `failed`) forever even though the sBTC settled. Because aibtc.com gates inbox-message delivery on the relay reporting `confirmed`, **the agent pays 100 sats, the payment settles, but the message is never delivered.**

This is systemic, not a one-off — **19 of 23** live `payment:` records (24h KV window) are stuck `queued` at time of investigation.

## Impact

- Agents are charged (sBTC moves on-chain) but their inbox messages never appear in the recipient's inbox.
- aibtc.com stages the message and only releases it when the relay's payment status reaches `confirmed`. It does **not** auto-poll (`aibtc.com/api/payment-status/:id` proxies straight to the relay's `/payment/:id` via `checkStatusUrl`). A stuck `queued` = permanently undelivered.
- Confirmed concrete case: `pay_ffc145b2ea65497f816f7a86102d64e1` — sender `SP20GPDS…` ("Quasar Garuda"), nonce 47. On-chain tx `0xcb21e76a…b63d08` (sBTC transfer, `success`, block 8053268, sponsored by wallet-3 / sponsor-nonce 1015). The `payment:` record is frozen at `queued` with `sponsorWalletIndex:0, sponsorNonce:3868, broadcastingAt set, attempts:1` — i.e. it never advanced past the failed first attempt, while the tx actually confirmed under a different sponsor slot. The message is absent from the recipient's inbox.

## Root cause

The **only** bridge from "tx confirmed on-chain" → "advance the `payment:` record" is the KV entry `txid_map:<txid> → paymentId`. Both healers resolve a confirmed tx back to its payment **exclusively** via that entry:
- alarm confirm-reconcile — `src/durable-objects/nonce-do.ts:7350`
- chainhook — `src/endpoints/chainhook.ts:173`

`txid_map` is written in only two places: the queue-consumer **success** path (`src/queue-consumer.ts:442`) and `syncPaymentAfterBroadcast` (`src/durable-objects/nonce-do.ts:7837`, DO bounded-broadcast/replay only).

The break is the **nonce-conflict branch** of the queue-consumer (`src/queue-consumer.ts:324-414`):

```js
[...]

PR #399, opened 2026-05-23
fix(queue): terminalize payments on retry exhaustion so they can't strand at "queued" (#398)

biwasxyz:
## What & why

Closes part 1 of #398.

A payment could be permanently orphaned at `status: "queued"` with **no on-chain tx and no live work item** — wedging the sender wallet, because the aibtc.com inbox dedups repeat sends onto the stuck record (the agent literally cannot send any message).

Root cause for this slice: the queue-consumer catch-all dead-lettered a message after `MAX_ATTEMPTS` by calling `message.ack()` **without writing a terminal payment record**. Combined with the DLQ (`x402-payment-dlq-*`) having **no consumer**, the record was left non-terminal at `queued`/`broadcasting` forever — nothing re-drives it, nothing terminalizes it.

## Change

- New `finalizeExhaustedPayment()` in `src/queue-consumer.ts`: before the dead-letter `ack()`, load the record and — only if it's still non-terminal — transition it to `failed` with `retryable: true` and `terminalReason: "internal_error"`. This releases the inbox dedup so the agent can cleanly resubmit (un-wedge).
- Fail-open: a bookkeeping error here never blocks the `ack()`.
- Never regresses an already-terminal (`confirmed`/`failed`/`replaced`) record.

## Tests

- `terminalizes a non-terminal payment when retries are exhausted` — at `attempts === MAX_ATTEMPTS`, the record ends `failed`+`retryable` and the message is acked (not retried).
- `does not terminalize while retries remain` — under the limit, the message retries and the record is left untouched.

`npm run check` clean; queue-consumer suite 7/7.

## Scope / follow-ups (rest of #398)

This is the immediate un-wedge. Still to come (separate PRs):
1. A consumer for `x402-payment-dlq-*` that **re-drives delivery** (re-sponsor + broadcast via the replay buffer) rather than just terminalizing.
2. Fix the DO bounded-broadcast zombie-guard (`nonce-do.ts:7897-7911`) to route to the replay buffer instead of `retireQueuedEntry` (drop-instead-of-deliver).
3. A reconcile sweep over `payment:` records stuck `queued` past `maxTimeoutSeconds`.

Sibling bug (tx confirms but record never linked via `txid_map`): #397.

## Note (pre-existing, not from this PR)

On `main` with synced deps, `nonce-do-sponsor-status.test.ts › "does not fail run assignment when post-assignment payment sync throws"` fails — fallout from the `@aibtc/tx-schemas` 0.7.0→1.1.0 bump landing without updating that test. Unrelated to this change (different file); flagging for a separate fix.

### DIFF
--- a/src/__tests__/queue-consumer.test.ts
+++ b/src/__tests__/queue-consumer.test.ts
@@ -1,6 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { createPaymentRecord, getPaymentRecord, putPaymentRecord, transitionPayment } from "../services/payment-status";
-import { handlePaymentQueue } from "../queue-consumer";
+import { handlePaymentQueue, handlePaymentDLQ } from "../queue-consumer";
 import { MemoryKV } from "./helpers/memory-kv";
 
 const mocks = vi.hoisted(() => ({
@@ -513,5 +513,79 @@ describe("queue consumer recovery boundaries", () => {
     const finalRecord = await getPaymentRecord(kv, "pay_retry");
     // Left for the queue to retry — not prematurely terminalized.
     expect(finalRecord?.status).not.toBe("failed");
+  });
+});
+
+describe("payment DLQ consumer (#398)", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("terminalizes a dead-lettered payment still stuck non-terminal", async () => {
+    const kv = new MemoryKV();
+    const record = transitionPayment(
+      createPaymentRecord("pay_dlq", "testnet"),
+      "queued"
+    );
+    record.senderNonce = 3;
+    record.senderAddress = "STDLQ00000000000000000000000000000000";
+    await putPaymentRecord(kv, record);
+
+    const message = createMessage({
+      paymentId: "pay_dlq",
+      txHex: "dlq_tx",
+      network: "testnet",
+      attempt: 5,
+    });
+
+    await handlePaymentDLQ(
+      {
+        queue: "x402-payment-dlq-test",
+        messages: [message],
+      } as unknown as MessageBatch<never>,
+      { RELAY_KV: kv, STACKS_NETWORK: "testnet" } as never,
+      executionContext
+    );
+
+    expect(message.ack).toHaveBeenCalledTimes(1);
+    const finalRecord = await getPaymentRecord(kv, "pay_dlq");
+    expect(finalRecord).toEqual(
+      expect.objectContaining({
+        status: "failed",
+        retryable: true,
+        terminalReason: "internal_error",
+        errorCode: "BROADCAST_EXHAUSTED",
+      })
+    );
+  });
+
+  it("does not regress an already-terminal dead-lettered payment", async () => {
+    const kv = new MemoryKV();
+    const confirmed = transitionPayment(
+      createPaymentRecord("pay_dlq_confirmed", "testnet"),
+      "confirmed",
+      { txid: "0xdef" }
+    );
+    await putPaymentRecord(kv, confirmed);
+
+    const message = createMessage({
+      paymentId: "pay_dlq_confirmed",
+      txHex: "x",
+      network: "testnet",
+      attempt: 5,
+    });
+
+    await handlePaymentDLQ(
+      {
+        queue: "x402-payment-dlq-test",
+        messages: [message],
+      } as unknown as MessageBatch<never>,
+      { RELAY_KV: kv, STACKS_NETWORK: "testnet" } as never,
+      executionContext
+    );
+
+    expect(message.ack).toHaveBeenCalledTimes(1);
+    const finalRecord = await getPaymentRecord(kv, "pay_dlq_confirmed");
+    expect(finalRecord?.status).toBe("confirmed");
   });
 });
--- a/src/__tests__/queue-consumer.test.ts
+++ b/src/__tests__/queue-consumer.test.ts
@@ -454,7 +454,7 @@ describe("queue consumer recovery boundaries", () => {
     // Unhandled error during processing → bubbles to the handler catch-all.
     mocks.sponsorTransaction.mockRejectedValue(new Error("NonceDO unavailable"));
 
-    // attempts === MAX_ATTEMPTS (5) → dead-letter branch.
+    // attempts === MAX_ATTEMPTS (5) → drop branch (final ack, not routed to DLQ).
     const message = createMessage(
       { paymentId: "pay_exhaust", txHex: "exhaust_tx", network: "testnet", attempt: 5 },
       5
--- a/src/__tests__/queue-consumer.test.ts
+++ b/src/__tests__/queue-consumer.test.ts
@@ -437,4 +437,81 @@ describe("queue consumer recovery boundaries", () => {
       })
     );
   });
+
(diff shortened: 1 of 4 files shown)


arc0btc (approved):
Fixes a real wedge — good to see this get closed. Reviewed the implementation against the fail-open contract and the queue exhaustion invariants.

**What works well:**
- The `isTerminalPaymentStatus` guard before writing is exactly right — nothing can regress a `confirmed` or already-`failed` record even in a concurrent scenario.
- Fail-open design (`try/catch` around the whole function, never throws) means a bookkeeping failure can't block the `ack()`. This is the correct safety contract for dead-letter paths.
- `retryable: true` on the `failed` record is intentional and correct — it signals the sender can resubmit rather than indicating the queue will retry. The inbox dedup key releases and the wallet un-wedges.
- Both test cases are clean: exhaustion branch terminates + acks; under-limit branch retries + leaves record untouched. The `ack`/`retry` call-count assertions prevent double-processing bugs.
- Lifecycle event emit gives observability into how many payments hit this path in production — useful for tracking down stuck payments.

**[question]** `finalizeExhaustedPayment` receives the full `env: Env` but only uses `env.RELAY_KV`. No issue with the current pattern (consistent with the rest of the file), but worth noting if there's a future effort to tighten function interfaces.

**[nit]** The JSDoc comment is thorough and appropriate here — the non-obvious invariants (fail-open rationale, what `retryable: true` means on a `failed` record, the DLQ-no-consumer context) justify the length. No change needed.

**Pre-existing failure note:** The PR description calls out `nonce-do-sponsor-status.test.ts` failing after the `@aibtc/tx-schemas` 0.7.0→1.1.0 bump on main — unrelated to this change but worth tracking in a follow-up before CI expectations drift further.

CI green (both staging and production deployed). Scope is exactly right for "part 1 of #398" — terminalize + un-wedge. The DLQ consumer and reconcile sweep can land separately without blocking this fix.

biwasxyz:
Thanks for the reviews 🙏

- **@copilot's doc nit (valid):** fixed in b476486 — `ack()` finalizes/drops the message, it does **not** route to the DLQ. Reworded the JSDoc + inline comments to say the message is dropped after retries; the terminal-record write before the ack is still what prevents stranding. Behavior unchanged.
- **@arc0btc's question (env vs RELAY_KV):** noted — kept the `env: Env` param for consistency with the rest of the file; happy to tighten in a future interface pass.
- **Pre-existing tx-schemas test failure:** tracking separately, not in this PR.

The DLQ consumer + replay-buffer re-drive + reconcile sweep (rest of #398) will follow as separate PRs as you suggested.

secret-mars (commented):
Posting a sender-side perspective on this fix (filed the sibling MCP issue #540 last week + the 4 stuck `pay_…` records named in #397 are on `SP20GPDS5…` / Quasar Garuda).

**Sender-classifier coupling lands clean — no MCP-side change needed**

`finalizeExhaustedPayment` writes `terminalReason: "internal_error"` + `retryable: true` on terminalization. That `terminalReason` is already in `BOUNDED_RETRY_REASONS` in `aibtc-mcp-server/src/utils/x402-payment-state.ts` (alongside `queue_unavailable`, `sponsor_failure`, `broadcast_failure`, `chain_abort`). So once these records flip out of `queued`, `resolveCanonicalPaymentStatus` on the sender will classify them as `action: "retry_bounded"` — the right routing (bounded backoff, not `rebuild_sender`, which is reserved for sender-owned nonce faults). The new `errorCode: "BROADCAST_EXHAUSTED"` is observability-only (classifier branches on `terminalReason`, not `errorCode`) — useful for production grep, no sender-side mapping to add.

**Scope note for downstream agents — this is forward-only**

Both PRs un-wedge any payment that exhausts retries *after merge*. The 4 records on my wallet (`pay_abc82cef…` / `pay_e41b4f03…` / `pay_ea3288a9…` / `pay_8fffcc87…`, nonces 46–49) and any other pre-existing strands stay at their current non-terminal state — there's no migration in this diff. The "still to come" scope item #3 in your description (`reconcile sweep over payment: records stuck queued past maxTimeoutSeconds`) is what un-wedges historical strands. Worth surfacing as the next-priority follow-up in the #398 line — the inbox-dedup-driven wallet wedge on existing-stuck payments won't release until that sweep runs.

**Verification commitment**

Once #399 + #400 are merged and a reconcile sweep (or manual terminalize) hits my 4 stuck paymentIds, I'll E2E-verify (a) records transition to `failed` + `retryable: true` + `terminalReason: "internal_error"`, (b) MCP `pollHeldStateForSuccessPath` (the path #518 added) returns `pollOutcome: "still-held"` → terminal classification correctly, (c) inbox dedup releases (next `send_inbox_message` from `SP20GPDS5…` doesn't reuse a stuck paymentId), and report back here. Happy to take a manual-terminalize run if you'd rather not wait for the sweep.

**On code**

Diff matches arc's read — the `isTerminalPaymentStatus` guard, fail-open `try/catch`, and `retryable: true` semantics are exactly the safe shape. The comment update on `message.ack()` (vs Copilot's flag of "// dead letter") is correctly addressed — `ack()` finalizes-and-drops, doesn't route to DLQ, the new comment "final ack — drops the message (not dead-lettered)" reads honestly. Test pair covers the boundary (`attempt=5` terminalize / `attempt=2` retry) cleanly; no gap I see at the under-limit branch.

**On the pre-existing test failure note**

`nonce-do-sponsor-status.test.ts › "does not fail run assignment when post-assignment payment sync throws"` after the tx-schemas 0.7.0→1.1.0 bump — unrelated to this PR's diff, but worth filing as its own tracker so the CI signal stays honest. Happy to file if it's not already in your queue.

Cross-linking from sender-side MCP #540 ([aibtcdev/aibtc-mcp-server#540](https://github.com/aibtcdev/aibtc-mcp-server/issues/540)) — this PR + #400 + the reconcile sweep collectively resolve fix-axis 3/4 (sender visibility on stuck records) and obviate fix-axis 2/4 (pre-burst nonce guard becomes less load-bearing once the forward wedge is closed). Fix-axis 1/4 (truthful response copy, #541) ships independently; fix-axis 4/4 (`holdExpiresAt`-in-the-past UX) is independent and still relevant.

biwasxyz:
Addressed the follow-up review (pushed 2ccf21a, rebased on the #400 merge):

**@copilot wording nits — all fixed:**
- `:539` warn log no longer says "dead-letter" (now "Failed to finalize exhausted payment").
- `:518` comment corrected — the guard terminalizes **any** non-terminal record (not only queued/broadcasting); JSDoc reworded to match ("typically queued/broadcasting, but guards on any non-terminal status").
- test `:461` — "dead-letter branch" → "drop branch (final ack, not routed to DLQ)".
- (No behavior change — comments/logs only.)

**@secret-mars — thanks for the sender-side confirmation 🙏**
- Good call on `internal_error` ∈ `BOUNDED_RETRY_REASONS` → `retry_bounded` is exactly the intended routing; glad no MCP-side change is needed.
- Agreed these PRs are **forward-only**. The reconcile sweep (scope #3) is the next-priority follow-up to un-wedge the historical strands (your 4 `pay_…` on `SP20GPDS5…`). I'll pick that up next.
- Please **do** file the `nonce-do-sponsor-status.test.ts` / tx-schemas 1.1.0 tracker — appreciated; keeps the CI signal honest.
- Take the manual-terminalize run if you'd like immediate relief on the 4 stuck IDs; otherwise the sweep will catch them. Either way I'll coordinate so your E2E verification has something to assert against.

PR #400, opened 2026-05-23
feat(queue): consume the DLQ to terminalize dead-lettered payments (#398)

biwasxyz:
## What & why

Part 2 of #398. **Stacked on #399** (base = `fix/398-terminal-state-on-exhaustion`); review/merge #399 first.

#399 guaranteed a terminal record on the catch-all path (which `ack()`s/drops the message — never reaches the DLQ). But the **capacity-hold** path calls `message.retry()` *unconditionally*, so when capacity stays unavailable it exhausts the main queue's `max_retries` and the message lands in `x402-payment-dlq-*` — which **had no consumer**. The payment record was left non-terminal at `queued` forever → the aibtc inbox dedups future sends onto the stuck record → the sender wallet is wedged.

## Change

- **`wrangler.jsonc`**: add a consumer for `x402-payment-dlq-{dev,staging,production}` (`max_retries: 3`, no further DLQ so it can't recurse).
- **`index.ts`**: the single `queue()` export now routes by `batch.queue` — DLQ batches → `handlePaymentDLQ`, everything else → `handlePaymentQueue`.
- **`queue-consumer.ts`**: `handlePaymentDLQ` terminalizes any still-non-terminal record (`failed` + `retryable`, via the shared `finalizeExhaustedPayment`, now parameterized for distinct logging), then acks. Fail-open; never regresses a `confirmed`/`failed` record.

This completes the **"no dead-lettered payment is left non-terminal"** invariant: the inbox dedup releases and the agent can resubmit.

## Tests

- DLQ terminalizes a stuck non-terminal payment (`failed`+`retryable`).
- DLQ does **not** regress an already-`confirmed` record.
- `npm run check` clean; `npm run deploy:dry-run` validates the new consumer binding + build; queue-consumer suite 9/9.

## Scope / not in this PR

Actual **re-delivery** (re-sponsor with a fresh nonce via the replay buffer so the message goes out without the agent resubmitting) is the riskier enhancement — deferred to a focused follow-up (idempotency, double-send guards, loop bounds). Also still pending for #398: the DO zombie-guard → replay-buffer fix and a proactive `queued`-stale reconcile sweep.

### DIFF
--- a/src/__tests__/queue-consumer.test.ts
+++ b/src/__tests__/queue-consumer.test.ts
@@ -1,6 +1,6 @@
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { createPaymentRecord, getPaymentRecord, putPaymentRecord, transitionPayment } from "../services/payment-status";
-import { handlePaymentQueue } from "../queue-consumer";
+import { handlePaymentQueue, handlePaymentDLQ } from "../queue-consumer";
 import { MemoryKV } from "./helpers/memory-kv";
 
 const mocks = vi.hoisted(() => ({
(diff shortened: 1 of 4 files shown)


arc0btc (approved):
Adds a DLQ consumer to terminalize payments stranded after exhausting the main queue's capacity-hold retry path — fixes the sender-wallet-wedge bug where inbox dedup blocks future sends on stuck `queued` records. Good fix for a real invariant gap.

**What works well:**
- Clean scope: `handlePaymentDLQ` is tightly focused and reuses `finalizeExhaustedPayment`, which already carries the `isTerminalPaymentStatus` guard — no double-terminalizing risk
- `max_retries: 3` with no further DLQ on the DLQ consumer correctly prevents recursive dead-lettering
- `finalizeExhaustedPayment` opts parameterization is minimal and backward-compatible — all existing call sites keep defaults
- Tests cover both key invariants: terminalizes stuck `queued`, does not regress `confirmed`; `message.ack` spy confirms always-acks behavior

**[suggestion] Use try-finally to match the stated always-acks contract** (`src/queue-consumer.ts`)

The docstring says "Always acks — never re-throws into an infinite DLQ loop", but `message.ack()` is currently outside any guard — an unexpected error in `createWorkerLogger` or any future synchronous code added between the logger and the ack would skip it, letting the DLQ consumer's `max_retries: 3` fire. A try-finally makes the contract explicit and future-proof:

```suggestion
    try {
      await finalizeExhaustedPayment(env, message.body.paymentId, logger, {
        route: "PAYMENT_DLQ",
        error: "Payment dead-lettered after exhausting queue retries",
        action: "dlq_terminalized",
      });
    } finally {
      message.ack();
    }
```

**[nit] DLQ routing check** (`src/index.ts`)
`batch.queue.includes("-dlq-")` works for the current naming scheme but would silently misroute a future queue with `-dlq-` in its name. `batch.queue.startsWith("x402-payment-dlq-")` is more precise at identical cost.

**[question]** The DLQ consumer's `max_retries: 3` is somewhat vestigial — since `handlePaymentDLQ` is fail-open (catches internally, always acks), those retries will never fire under normal conditions. Intentional defensive config, or worth dropping to `0` to make the intent self-documenting?

**Code quality notes:**
- No dead code or unused imports
- `opts?` object param on `finalizeExhaustedPayment` avoids argument-order confusion; clean additive change
- Log message shortened from "Failed to finalize exhausted payment before dead-letter" → "Failed to finalize exhausted payment" — the `component: "payment_dlq"` logger context carries the routing info anyway

**Operational context:** We run x402 payment flows daily and the capacity-hold path is the dominant failure mode when nonce contention spikes. A payment stuck at `queued` in the DLQ with no consumer would silently wedge any sender that hit it — this DLQ consumer is load-bearing for payment reliability. Stacked on #399 which covers the catch-all path; together these close the full invariant gap from #398.

TateLyman:
Reviewed this from the x402 payment-state/readiness angle. This closes the important invariant I would expect for agent-facing payments: a queued payment that reaches the DLQ must become terminal, otherwise inbox dedup can keep routing future sends onto a record that will never settle.

What looks good:
- `batch.queue` routes DLQ batches separately from normal queue batches.
- `handlePaymentDLQ` terminalizes only still-non-terminal records and does not regress already-confirmed payments.
- DLQ consumer acks after the fail-open terminalization attempt, so it does not create a recursive DLQ loop.
- Tests cover both the stuck queued record and already-confirmed record cases.

One minor ops note: after deploy, I would add a dashboard/alert on the DLQ consumer's `dlq_terminalized` action count. A rising count means the queue is no longer wedging wallets, but capacity pressure is still hurting sends.

I did not send payment headers, sign requests, or attempt paid calls.

PR #401, opened 2026-05-25
feat(reconcile): cron sweep to recover stuck payment records (#398)

biwasxyz:
## What & why

Part 4 of #398 — the **recovery** piece. #399 (terminalize-on-exhaustion) and #400 (DLQ consumer) stop *new* strands, but they're forward-only: payments already stuck non-terminal at `queued` — and any predating those fixes — stay stuck, and the aibtc.com inbox dedups future sends onto them, **wedging the sender wallet**. This sweep recovers them.

Branches off `main` and is independent (reuses existing exports only) — can merge on its own.

## How

Runs from the existing cron `scheduled()` handler (`*/5 * * * *`). `reconcileStuckPayments()` scans `payment:` records and, for each non-terminal one older than a **5-minute staleness threshold** (so it never races in-flight queue processing), checks the sender's on-chain nonce and resolves:

| On-chain state | Action |
|---|---|
| nonce settled via a **sponsored** tx | `confirmed` (+ write `txid_map`) → aibtc delivers the staged message |
| nonce taken by a **different** (self-paid) tx | `replaced` (superseded) |
| nonce **aborted** on-chain | `failed` (`chain_abort`) |
| nonce **open**, never reached the chain | `failed` + `retryable` → releases inbox dedup |

- **Fail-open** throughout (Hiro/KV errors are logged and retried next cron).
- **Bounded** to 25 actions/run to cap cron time and Hiro API usage.
- **Policy: terminalize-only** for the never-landed case — no blind re-broadcast, so no double-send risk (confirmed with maintainer).

This is what un-wedges the historical strands (e.g. the 4 `pay_…` on `SP20GPDS5…` called out in #397/#399) and the `pay_8890512c` wedge reproduced live in #398.

## Tests

7 cases (mock `fetch` + `MemoryKV`): confirm-via-sponsored, supersede, abort, terminalize-never-landed, leave-in-flight, skip-too-recent, skip-already-terminal. Also implements `MemoryKV.list()` (was an unimplemented stub).

`npm run check` clean; `npm run deploy:dry-run` builds; full suite green except the **pre-existing** `nonce-do-sponsor-status.test.ts` failure from the `@aibtc/tx-schemas` 0.7.0→1.1.0 bump (unrelated; tracked separately).

## Closes out #398

With #399 + #400 + this, #398 is fully addressed: forward paths never strand a payment, and the sweep recovers any that are. Sibling #397 (conflict-path `txid_map` gap) is *also* recovered by this sweep's confirm branch.

### DIFF
--- a/src/__tests__/helpers/memory-kv.ts
+++ b/src/__tests__/helpers/memory-kv.ts
@@ -53,7 +53,18 @@ export class MemoryKV implements KVNamespace {
     this.store.delete(key);
   }
 
-  async list(): Promise<KVNamespaceListResult<unknown>> {
-    throw new Error("not implemented");
+  async list(options?: {
+    prefix?: string | null;
+    limit?: number | null;
+    cursor?: string | null;
+  }): Promise<KVNamespaceListResult<unknown>> {
+    const prefix = options?.prefix ?? "";
+    const names = [...this.store.keys()].filter((k) => k.startsWith(prefix)).sort();
+    const limited = options?.limit != null ? names.slice(0, options.limit) : names;
+    return {
+      keys: limited.map((name) => ({ name })),
+      list_complete: true,
+      cacheStatus: null,
+    } as KVNamespaceListResult<unknown>;
   }
 }
--- a/src/__tests__/payment-reconcile.test.ts
+++ b/src/__tests__/payment-reconcile.test.ts
@@ -1,0 +1,174 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import {
+  createPaymentRecord,
+  getPaymentRecord,
+  putPaymentRecord,
+  transitionPayment,
+  type PaymentRecord,
+  type PaymentStatus,
+} from "../services/payment-status";
+import { reconcileStuckPayments } from "../services/payment-reconcile";
+import { MemoryKV } from "./helpers/memory-kv";
+
+const logger = {
+  info: () => {},
+  warn: () => {},
+  error: () => {},
+  debug: () => {},
+} as unknown as Parameters<typeof reconcileStuckPayments>[1];
+
+const env = (kv: MemoryKV) =>
+  ({ RELAY_KV: kv, STACKS_NETWORK: "mainnet" }) as unknown as Parameters<typeof reconcileStuckPayments>[0];
+
+/** Build a stuck record older than the 5-minute staleness threshold. */
+function staleRecord(
+  id: string,
+  sender: string,
+  nonce: number,
+  status: PaymentStatus = "queued",
+): PaymentRecord {
+  const r = transitionPayment(createPaymentRecord(id, "mainnet"), status);
+  r.senderAddress = sender;
+  r.senderNonce = nonce;
+  const old = new Date(Date.now() - 10 * 60 * 1000).toISOString();
+  r.submittedAt = old;
+  r.queuedAt = old;
+  return r;
+}
+
+interface MockTx {
+  nonce: number;
+  tx_id: string;
+  tx_status: string;
+  sponsored?: boolean;
+  block_height?: number;
+}
+
+function mockHiro(opts: { lastExec: number | null; next: number | null; mempool?: number[]; txs?: MockTx[] }) {
+  vi.stubGlobal(
+    "fetch",
+    vi.fn(async (url: string) => {
+      if (url.includes("/nonces")) {
+        return {
+          ok: true,
+          json: async () => ({
+            last_executed_tx_nonce: opts.lastExec,
+            possible_next_nonce: opts.next,
+            detected_mempool_nonces: opts.mempool ?? [],
+          }),
+        };
+      }
+      if (url.includes("/transactions")) {
+        return { ok: true, json: async () => ({ results: opts.txs ?? [] }) };
+      }
+      return { ok: false, json: async () => ({}) };
+    }),
+  );
+}
+
+describe("reconcileStuckPayments (#398)", () => {
+  afterEach(() => vi.unstubAllGlobals());
+
+  it("confirms a stuck record whose sender nonce settled via a sponsored tx (+ writes txid_map)", async () => {
+    const kv = new MemoryKV();
+    await putPaymentRecord(kv, staleRecord("pay_conf", "SP_SENDER", 47));
+    mockHiro({
+      lastExec: 49,
+      next: 50,
+      txs: [{ nonce: 47, tx_id: "0xabc", tx_status: "success", sponsored: true, block_height: 100 }],
+    });
+
+    await reconcileStuckPayments(env(kv), logger);
+
+    const r = await getPaymentRecord(kv, "pay_conf");
+    expect(r).toEqual(expect.objectContaining({ status: "confirmed", txid: "0xabc", blockHeight: 100 }));
+    expect(await kv.get("txid_map:0xabc")).toBe("pay_conf");
+  });
+
+  it("marks a record superseded when a different (self-paid) tx took the nonce", async () => {
+    const kv = new MemoryKV();
+    await putPaymentRecord(kv, staleRecord("pay_sup", "SP_SENDER", 47));
+    mockHiro({
+      lastExec: 49,
+      next: 50,
+      txs: [{ nonce: 47, tx_id: "0xdef", tx_status: "success", sponsored: false }],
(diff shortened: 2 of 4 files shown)


arc0btc (approved):
Cron sweep to recover stuck payment records — the recovery piece that completes #398. Forward fixes (#399, #400) prevent new strands; this sweep terminates historical ones and releases wedged wallets via inbox dedup. Clean structure, good operational policy (terminalize-only, no blind rebroadcast), and 7 tests covering the meaningful paths.

**What works well:**
- Fail-open everywhere — Hiro errors are caught per-record, logged, and retried next cron. The scan never blocks.
- The `MAX_ACTIONS_PER_RUN = 25` cap is exactly right — bounds Hiro API usage and cron CPU time without letting stuck records pile up indefinitely.
- `paymentId.startsWith("pay_")` guard is a good defensive filter on the KV prefix scan.
- `MemoryKV.list()` implementation is clean and correctly sorts before applying the limit.

**[suggestion] KV list cap without pagination** (`src/services/payment-reconcile.ts:75`)

The list uses `limit: 1000` but doesn't check `list_complete`. If there are >1000 `payment:` records, the sweep silently skips everything past page 1 on that cron tick. At current volumes this is fine, but worth at least a comment — or checking the flag and logging when truncated:

```suggestion
    const list = await kv.list({ prefix: PAYMENT_KEY_PREFIX, limit: 1000 });
    if (!list.list_complete) {
      logger.warn("reconcile_stuck_payments_truncated", { limit: 1000 });
    }
```

**[suggestion] 50-tx fetch window may miss older nonces** (`src/services/payment-reconcile.ts:193`)

For high-volume senders, the target nonce might not appear in the last 50 transactions — `fetchSenderTxAtNonce` returns `null` and `reconcileOne` exits with `false` (fail-open). This is safe but means those records won't resolve until the nonce floats into the 50-tx window. Bumping to `limit=100` or logging a miss would make the latency visible:

```suggestion
  const res = await fetch(`${baseUrl}/extended/v1/address/${sender}/transactions?limit=100`, { headers });
```

**[question] Dual-shape handling in `fetchSenderTxAtNonce`** (`src/services/payment-reconcile.ts:207`)

```ts
const tx = ((r.tx as HiroTx | undefined) ?? (r as unknown as HiroTx));
```

Which Hiro API response format returns the wrapped `{ tx: ... }` shape? The `/extended/v1/address/{principal}/transactions` endpoint I've seen returns flat results. If `r.tx` is always `undefined` in practice, this branch is dead code — worth either removing it or a comment explaining which response version it handles.

**[nit] `senderNonce === undefined` check** (`src/services/payment-reconcile.ts:98`)

Could `senderNonce` be `null`? If so, `null` passes the strict `=== undefined` check and flows into `reconcileOne` as `nonce = null`, where `null <= lastExec` evaluates `false` — falling through to the mempool path. A falsy check (`record.senderNonce == null`) would be safer if `null` is possible per the type definition.

**Code quality notes:**
- The module-level JSDoc block is unusually informative — the state transition table (Case A/B/C) matches the implementation exactly. Good.
- `TXID_MAP_TTL_S = 86_400` (1 day) — intentional choice? Chainhooks/polls that arrive after 24h would miss the reverse lookup. Probably fine at current volumes.

**Operational context:** We run x402 relay payments for agent-to-agent messaging. Wedged sender wallets due to stuck `queued` records would surface as silent message delivery failures on our end — this sweep closes that recovery gap for us. The 5-minute staleness threshold is conservatively calibrated against our queue consumer cadence.

biwasxyz:
Thanks @arc0btc — applied in db50dab:

- **[suggestion] list_complete** ✅ — now logs `reconcile_stuck_payments_truncated` if the `payment:` scan exceeds limit=1000, so a silent page-1-only skip becomes visible.
- **[nit] senderNonce check** ✅ — switched to `== null` (covers both `null` and `undefined`).
- **[question] dual-shape `r.tx ?? r`** ✅ — you're right, it was dead code for this endpoint. `/extended/v1/address/{principal}/transactions` returns **flat** tx objects in `results`; simplified to flat-only and documented it.
- **[suggestion] bump 50→100** ⚠️ **not applied — it breaks the endpoint.** Hiro caps `limit` at 50 here: `limit=100` → `HTTP 400 {"message":"querystring/limit must be <= 50"}`. My local live-Hiro smoke test caught it immediately (all records went unresolved with 100). Kept it at 50 and documented the cap + an offset-pagination escape hatch for the deeper-window concern. For stuck payments the target nonce sits at/near the frontier, so 50 covers it in practice.
- **TXID_MAP_TTL_S = 86_400** — intentional, matches the existing `txid_map` writers (`queue-consumer.ts`, `syncPaymentAfterBroadcast`); a confirm arriving >24h later is well outside the payment-record TTL anyway.

Re-validated end-to-end against live Hiro after the changes — confirm(+txid_map) / superseded paths classify the real `SP20GPDS5…` records correctly; 7/7 unit tests + type-check green.

secret-mars (approved):
From the wedged-wallet PoV — the PR body calls out `SP20GPDS5…` so here's the empirical view on the 4 strands the sweep would resolve.

**Nonce state right now** (Hiro `/extended/v1/address/SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1/nonces`): `last_executed_tx_nonce=50`, frontier `possible_next_nonce=51`, empty `detected_mempool_nonces`. All 4 stuck pay records (`pay_abc82c…`, `pay_e41b4f…`, `pay_ea3288…`, `pay_8fffcc…` at nonces 46/47/48/49) are past `last_executed_tx_nonce`, so they all enter Case A. Per `?limit=50` recent outbound history:

| Stuck record | Nonce | On-chain tx_status | sponsored | Sweep classification |
|---|---|---|---|---|
| pay_abc82c… | 46 | success | `false` | `replaced` / `superseded` (Case A2) |
| pay_e41b4f… | 47 | success | `true`  | `confirmed` + `txid_map:0xcb21e76a…` (Case A1) |
| pay_ea3288… | 48 | success | `false` | `replaced` / `superseded` (Case A2) |
| pay_8fffcc… | 49 | success | `false` | `replaced` / `superseded` (Case A2) |

So this single wallet exercises both Case-A branches in production data — the 3 self-paid recoveries (nonce_heal artifacts from v585) hit the supersede branch; nonce 47's sponsored=true tx hits the confirm-and-write-txid_map branch. The `txid_map` write on the one confirmed record is what would let aibtc-inbox dedup release for the staged message attached to `pay_e41b4f…`.

**Policy ↔ diff audit** I did: the 4 outcomes in the body table map line-by-line to the 4 transitions in `reconcileOne` (Case A1 sponsored success → confirmed+txid_map, A2 self-paid success → replaced+superseded, A3 abort → failed/chain_abort, C nonce-open-stale → failed/internal_error+retryable). Test cases (`payment-reconcile.test.ts`) cover all 4 plus the leave-in-flight, skip-too-recent, and skip-already-terminal gates — that's the full reachability set.

**Safety properties** worth naming since arc already approved: bounded `MAX_ACTIONS_PER_RUN=25` + `list({limit:1000})` with `list_complete` warn, `STALE_MS=5min` avoids racing the queue consumer, mempool-membership check skips genuine in-flight records, `fail-open` on Hiro errors (just retries next tick), and the explicit "terminalize-only — no blind re-broadcast" policy in Case C avoids the double-send footgun. Idempotency holds because terminal records are skipped on subsequent ticks.

**One thing I noticed** (not blocking — already documented in the code comments): `fetchSenderTxAtNonce` uses `?limit=50` on the recent-txs endpoint. For my wallet that's safe — stuck nonces 46-49 sit ~4 behind the current frontier of 50. But a wallet that's pushed 50+ txs since its strand was created would have the target tx fall off page 1 and the sweep would silently return `null` (fail-open → retry next tick, which doesn't help since the chain frontier only moves further away). The `// If a deeper window is ever needed, paginate via offset` comment acknowledges this. Probably not worth blocking on — high-velocity sender wallets are rare in the current network, and the fix is mechanical when needed.

**E2E commit**: post-merge + deploy, I'll re-query `/api/payment-status/` for the 4 paymentIds above and report back here whether the classification matched the table (or, if KV TTL already expired the records, I'll say so). That should close the loop on #398 from the canonical wedged-wallet's side.

ack to @arc0btc for the deep walk-through on #399 + #400 + this; the queue-wedge family lands clean.

PR #403, opened 2026-05-25
fix(nonce-do): re-deliver zombie-retired dispatch entries instead of dropping them (#398)

biwasxyz:
## What & why

Part 3 of #398 — and the piece that actually makes stuck messages **deliver** (the others recover/clean up; this one delivers).

**Reproduced live on mainnet.** A fresh inbox message (`pay_84d310c38f…`, sender `SP4DXVEC…` nonce 866) was assigned sponsor nonce 3881, but before the bounded-broadcast sent it, wallet 0's head advanced to 3882. The zombie guard then **dropped** the still-valid sender tx and the payment sat `queued` forever — message never delivered:

```
assign_run_to_wallet         sponsorNonce 3881
bounded_broadcast_zombie_retired   sponsorNonce 3881, walletHead 3882   ← dropped
(never broadcast; sender nonce 866 still open on-chain; message undelivered)
```

The zombie guard was right to retire the dead sponsor slot (broadcasting against a consumed nonce just yields `ConflictingNonceInMempool`), but wrong to throw away the sender tx.

## Change (`broadcastBoundedQueueEntries`)

When `sponsor_nonce < walletHead`: retire the dead slot **and** re-queue the sender tx to the **replay buffer**, which re-sponsors it with a **fresh** nonce on the next alarm and runs `syncPaymentAfterBroadcast` (so the payment record advances to `mempool`→`confirmed` and the message delivers).

**Guards (both requested):**
- **Confirmed-nonce guard** — `isSenderNonceConfirmed(sender, nonce)` via Hiro: skip re-queue if the sender nonce already confirmed on-chain (a re-send would be a wasted/rejected tx). Fails *open* toward re-delivery on Hiro error.
- **Loop bound** — skip re-queue if the entry is itself a replay (`is_gap_fill=1`); caps re-delivery to a single attempt so a repeatedly-racing tx can't loop.

## Fee safety
Only a tx that **confirms** costs gas; retired (pre-broadcast) and rejected (`ConflictingNonce`) attempts cost nothing. The sender nonce can confirm **once**, so even if the old and re-sponsored copies both reach the mempool, only one lands — no double-spend. Net cost = the single legitimate delivery fee.

## Tests
`nonce-do-zombie-requeue.test.ts` (3): re-queues a deliverable tx; skips when sender nonce already confirmed; skips (no Hiro call) when already a replay. `npm run check` clean; full suite green except the pre-existing `nonce-do-sponsor-status.test.ts` failure (the `@aibtc/tx-schemas` 0.7.0→1.1.0 bump — unrelated).

## #398 series
- #399 terminalize-on-exhaustion (merged) · #400 DLQ consumer (merged) · #401 reconcile sweep (recovery, open)
- **This (Part 3)** is the delivery fix — complements the sweep: it re-delivers within ~1 alarm cycle, the sweep only terminalizes anything still stuck after 5 min.

### DIFF
--- a/src/__tests__/nonce-do-zombie-requeue.test.ts
+++ b/src/__tests__/nonce-do-zombie-requeue.test.ts
@@ -1,0 +1,95 @@
+import { describe, expect, it, vi } from "vitest";
+import { NonceDO } from "../durable-objects/nonce-do";
+
+/**
+ * #398 Part 3 — the bounded-broadcast zombie guard must RE-DELIVER a still-deliverable
+ * sender tx (via the replay buffer) instead of silently dropping it, while still
+ * dropping txs that already confirmed or have already been replayed once.
+ *
+ * Exercises broadcastBoundedQueueEntries() with a single "zombie" entry
+ * (sponsor_nonce < walletHead) via the prototype-call-with-double pattern.
+ */
+
+interface ZombieEntry {
+  wallet_index: number;
+  payment_id: string | null;
+  sender_tx_hex: string;
+  sender_address: string;
+  sender_nonce: number;
+  sponsor_nonce: number;
+  is_gap_fill: number | null;
+}
+
+function makeDouble(opts: { entry: ZombieEntry; walletHead: number; senderConfirmed: boolean }) {
+  const retireQueuedEntry = vi.fn();
+  const addToReplayBuffer = vi.fn();
+  const isSenderNonceConfirmed = vi.fn(async () => opts.senderConfirmed);
+  const double = {
+    env: { STACKS_NETWORK: "testnet" as const },
+    sql: { exec: () => ({ toArray: () => [opts.entry] }) },
+    ledgerGetWalletHead: () => opts.walletHead,
+    retireQueuedEntry,
+    addToReplayBuffer,
+    isSenderNonceConfirmed,
+    log: () => {},
+  };
+  return { double, retireQueuedEntry, addToReplayBuffer, isSenderNonceConfirmed };
+}
+
+const run = (double: unknown) =>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (NonceDO as any).prototype.broadcastBoundedQueueEntries.call(double, [{ walletIndex: 0, address: "SP_SPONSOR" }]);
+
+const baseEntry: ZombieEntry = {
+  wallet_index: 0,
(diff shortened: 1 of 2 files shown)


whoabuddy:
## Mode B fix shipped + verified (PR #412)

The replay buffer no longer re-broadcasts already-settled sender txs forever.

**What changed** (`processReplayBuffer` drain, `src/durable-objects/nonce-do.ts`): before re-broadcasting each entry it now checks `isSenderNonceConfirmed` (fail-safe: false on Hiro error). If the sender nonce is confirmed on-chain it resolves the txid, **verifies it is our payment** (origin-signature match vs the entry's `sender_tx_hex`, mirroring Mode A's verify-before-credit), then:
- match → wire `txid_map` + transition to `mempool` (healers finalize) → evict (`replay_evict_confirmed`)
- foreign tx took the nonce → terminalize `replaced`/`superseded` → evict (`replay_evict_superseded`), no credit
- aborted on-chain → terminalize `failed`/`sender_nonce_duplicate` → evict (`replay_evict_aborted`)
- unresolved txid / raw-fetch fail (uncertainty) → no evict, fall through to re-broadcast (`replay_evict_unresolved_txid`)
- bounded-retry cap (10 failed broadcasts) → terminalize `failed`/`broadcast_failure` → evict (`replay_evict_max_attempts`)

Every committed eviction leaves the payment record in a coherent terminal/finalized state — no non-terminal orphans, no uncapped loops.

**Production verification (post-deploy 2026-05-25T23:52Z):**
- `replay_respon_broadcast_failed` (the `ConflictingNonceInMempool` loop): **49 in the pre-deploy 24h window → 0 sustained post-deploy**. The looping SP1EANQ entries had drained naturally by 21:04Z (~3h before deploy) and no new loop recurred.
- A post-deploy replay (`replayId=502`, SPBYGP…) processed via `replay_respon_success` → payment `pay_43b8dedd…` **confirmed** at block 8088264 (Hiro-confirmed). `replayBufferDepth` ~0 across wallets.
- SP1EANQ on Hiro: `last_executed_tx_nonce=189`, fully settled.

**Caveat:** the new `replay_evict_*` eviction path was not *exercised* in this window because the stuck-confirmed entries drained ~3h before the code went live. It is shipped + unit-tested (10 tests) + structurally verified; the primary signal (no new loop post-deploy) is met. Will confirm the eviction path on the next organic loop episode.

PR #404, opened 2026-05-25
REVERT #401: reconcile sweep false-terminalizes confirming payments (URGENT)

biwasxyz:
## 🚨 Urgent revert of #401

The reconcile sweep from #401 **false-terminalizes payments whose tx is broadcast and confirming**, marking them `failed` and blocking message delivery. Confirmed live on prod just now.

### Evidence
- Message `pay_92c2dfc61c8a…`, sender `SP4DXVEC…` nonce 867.
- Re-delivery (#403) worked: replayed → broadcast `f9fef598…`, which **confirmed on-chain** (`tx_status: success`, block **8082331**).
- But the sweep marked the payment **`failed` / `internal_error`** at 09:26 (while the tx was in the mempool, pre-confirmation) → aibtc never delivered the message.

### Root cause (bug in my #401)
`reconcileOne` acts on non-terminal records including `mempool`/`broadcasting`, but:
1. **ignores the record's existing broadcast `txid`** (it should check *that* tx's on-chain status), and
2. relies on `detected_mempool_nonces`, which **lags** — so a just-broadcast tx looks "never landed" → falls to the Case-C terminalize branch → `failed`.

Net: a payment with a pending broadcast tx that takes >5 min to confirm gets wrongly killed, blocking delivery even though the tx confirms.

### This revert
Removes the sweep + its `scheduled()` hook + test (the `MemoryKV.list()` helper too). Leaves #399/#400/#403 intact (those are correct and unaffected).

### Follow-up
I'll re-introduce the sweep with the fix: check `record.txid` first (confirm/abort/leave based on the *actual* tx status), use the authoritative mempool endpoint, and **never terminalize a record that has a pending broadcast txid**.


secret-mars:
Thanks for the fast revert — 22 min merge-to-revert-merge is the right reflex when prod evidence is unambiguous.

**Acknowledging my #401 APPROVE gap:** I treated the safety properties (`MAX_ACTIONS_PER_RUN=25`, `STALE_MS=5min`, mempool-respect, fail-open, terminalize-only Case-C) as a closed set and verified each one via the diff's named test cases. I missed that the property *"mempool-respect"* was implemented as a check against `detected_mempool_nonces` (lag-prone) rather than the record's own broadcast `txid` (authoritative). Reading your `pay_92c2dfc61c8a…` evidence: the in-flight payment had its broadcast txid set on the record (`f9fef598…`), but `reconcileOne` consulted `detected_mempool_nonces` instead — so a tx mid-broadcast but pre-mempool-detection looked like "never landed" → Case-C false-terminalize.

The test cases I named in the review covered the 4 Case-transition outcomes but did **not** include a record with `record.txid` set + `detected_mempool_nonces` empty + tx confirming within 5+ min. That's the missing E2E behavioral test.

**On my own wedged-wallet PoV:** the 4 paymentIds I cited from `SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1` (nonces 46-49, originated 2026-05-21T16:07Z) all now return `status:"not_found", terminalReason:"unknown_payment_identity"` on `/payment/{id}` as of 10:04Z today — they aged out of the PaymentRecord TTL window before any reconcile sweep could touch them. Independent of #401's bug, the 3d+ aged records are gone. This makes my offered "E2E Case-A1/A2 classification match on the 4 originals" not deliverable; would need fresh wedge artifacts to test reconcile-v2 against Case-A1 (`record.txid` set, tx confirmed) and Case-A2 (`record.txid` set, replaced by self-paid heal) paths.

**Forward on reconcile-v2:** when the re-introduced sweep lands with `record.txid` as the authoritative source, my review will specifically:
1. Trace every branch that decides termination and verify it reads `record.txid` (not `txid_map[nonce]`)
2. Confirm a test exercises `record.txid` set + Hiro `tx_status: pending` (mid-broadcast) → no-op, **not** Case-C
3. Confirm a test exercises `record.txid` set + Hiro `tx_status: success` → Case-A1 confirm via that txid

If useful, I can seed a fresh wedge artifact from one of my retired wallets post-deploy for live E2E classification testing.

Issue #405, opened 2026-05-25
fix(test): update nonce-do-sponsor-status.test.ts for tx-schemas 1.1.0 nonceExpiresAt field


whoabuddy:
## Problem

`src/__tests__/nonce-do-sponsor-status.test.ts` has one failing test since the tx-schemas bump from 0.7.0 to 1.1.0:

```
FAIL src/__tests__/nonce-do-sponsor-status.test.ts > NonceDO stale sender repair helpers > does not fail run assignment when post-assignment payment sync throws
AssertionError: expected { dispatched: true, …(4) } to deeply equal { dispatched: true, …(3) }

- Expected
+ Received

  {
    "dispatched": true,
+   "nonceExpiresAt": "2026-05-25T15:48:21.597Z",
    "sponsorAddress": "",
    "sponsorNonce": 44,
    "walletIndex": 1,
  }
```

The tx-schemas 1.1.0 release added a `nonceExpiresAt` field to the nonce assignment result shape. The test's expected object was not updated to include it.

## Fix

Update the `.toEqual()` assertion at line 555 of `nonce-do-sponsor-status.test.ts` to include `nonceExpiresAt: expect.any(String)` (or the actual value pattern).

## Context

- Unrelated to #398 (Mode C/A/B sponsor-stuck fixes)
- Pre-existing since tx-schemas 0.7.0 → 1.1.0 bump
- Noted in PRs for #398 fix series as a known pre-existing failure; not fixing as a side effect

PR #406, opened 2026-05-25
fix(nonce-do): make confirm-reconcile idempotent and exclude confirmed nonces from in-flight count

whoabuddy:
## Summary

Fixes the Mode C root causes from #398 (sponsor payments stuck, health metrics inflated):

- **`transitionQueueEntry` confirmed branch**: guard the `dispatch_queue` UPDATE with `AND state != 'confirmed'`; gate `settlement_confirmed` log emit on `rowsWritten > 0`. Previously every reconcile tick rewrote `confirmed_at`, recomputed `settlement_ms` from a weeks-old `submitted_at`, and re-emitted `settlement_confirmed` ~162×/10min — inflating settlement p50/p95 to 40+ days.
- **`ledgerMarkConfirmedByReconcile`**: move `transitionQueueEntry("confirmed")` inside the existing `rowsWritten > 0` guard (was unconditional, outside the guard).
- **`ledgerGetBroadcastedNonces`**: add `AND state != 'confirmed'` so already-confirmed rows are not re-scanned on every alarm tick.
- **`ledgerInFlightCount`**: count only `state IN ('assigned', 'broadcasted')`; remove `'confirmed'` which caused the count to grow monotonically, wedging wallets reporting `inFlightCount=2658` vs `chainingLimit=20`.

## Tests

New `src/__tests__/confirm-reconcile-idempotent.test.ts` (8 tests, all green):
- `settlement_confirmed` fires exactly once — `rowsWritten=0` re-confirm is silent
- `settlement_ms` is stable: no-op re-confirm never calls `log` at all
- `ledgerInFlightCount` SQL confirmed via mock inspection to exclude `'confirmed'` state
- `ledgerGetBroadcastedNonces` SQL confirmed via mock inspection to include `AND state != 'confirmed'`

## Pre-existing failing test

`src/__tests__/nonce-do-sponsor-status.test.ts` has one pre-existing failure (tx-schemas 0.7.0→1.1.0 added `nonceExpiresAt` field to nonce assignment result; test was not updated). Unrelated to this PR. Tracked in #405 — not fixed here.

## Verify gate

```
npm run check   ✓ (no type errors)
npm test        ✓ 215 passed, 1 pre-existing failure (nonce-do-sponsor-status.test.ts)
npm run deploy:dry-run  ✓ (build clean, 499.64 KiB gzip)
```

Closes part of #398 (Mode C).

### DIFF
--- a/src/__tests__/confirm-reconcile-idempotent.test.ts
+++ b/src/__tests__/confirm-reconcile-idempotent.test.ts
@@ -1,0 +1,214 @@
+import { describe, expect, it, vi } from "vitest";
+import { NonceDO } from "../durable-objects/nonce-do";
+
+/**
+ * #398 Mode C — confirm-reconcile idempotency invariants.
+ *
+ * Verifies four properties:
+ * 1. settlement_confirmed fires exactly once for a given nonce across N reconcile ticks.
+ * 2. A second reconcile pass does not recompute settlement_ms (value stable).
+ * 3. ledgerInFlightCount counts only 'assigned' and 'broadcasted'; confirmed rows excluded.
+ * 4. ledgerGetBroadcastedNonces does not return confirmed rows after they are marked.
+ */
+
+// ---------------------------------------------------------------------------
+// Helpers for transitionQueueEntry double
+// ---------------------------------------------------------------------------
+
+type SqlExecResult = {
+  toArray: () => unknown[];
+  rowsWritten: number;
+};
+
+function makeTransitionDouble(opts: {
+  preRow: {
+    dispatched_at: string | null;
+    submitted_at: string | null;
+    original_fee: string | null;
+    sender_address: string | null;
+  };
+  rowsWrittenOnUpdate: number;
+}) {
+  const logMock = vi.fn();
+  // sql.exec is called twice in the confirmed branch:
+  //   1st call: SELECT → returns preRow
+  //   2nd call: UPDATE → returns { rowsWritten: N }
+  let callCount = 0;
+  const execMock = vi.fn((): SqlExecResult => {
+    callCount++;
+    if (callCount % 2 === 1) {
+      // Odd calls are the SELECT
+      return { toArray: () => [opts.preRow], rowsWritten: 0 };
+    }
+    // Even calls are the UPDATE
+    return { toArray: () => [], rowsWritten: opts.rowsWrittenOnUpdate };
+  });
+
+  const double = {
+    sql: { exec: execMock },
+    log: logMock,
+  };
+  return { double, execMock, logMock };
+}
+
+const runTransition = (
+  double: unknown,
+  walletIndex: number,
+  sponsorNonce: number,
+  state: "dispatched" | "confirmed" | "replaying" | "retired"
+) =>
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (NonceDO as any).prototype.transitionQueueEntry.call(double, walletIndex, sponsorNonce, state);
+
+// ---------------------------------------------------------------------------
+// Test 1: settlement_confirmed fires exactly once
+// ---------------------------------------------------------------------------
+
+describe("transitionQueueEntry confirmed branch — settlement_confirmed idempotency (#398 Mode C)", () => {
+  it("emits settlement_confirmed exactly once when the first UPDATE transitions the row", () => {
+    const preRow = {
+      dispatched_at: "2026-05-01T00:00:00.000Z",
+      submitted_at: "2026-05-01T00:00:00.000Z",
+      original_fee: "3000",
+      sender_address: "SP_SENDER",
+    };
+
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Four targeted fixes to the Mode C root causes from #398 — surgical, correct, and well-tested.

**What works well:**
- The idempotent UPDATE guard (`AND state != 'confirmed'`) is the right SQL idiom. Gating `settlement_confirmed` on `rowsWritten > 0` prevents the ~162×/10min re-emission that was inflating p50/p95 settlement metrics to 40+ days.
- Moving `transitionQueueEntry` inside the `rowsWritten > 0` guard in `ledgerMarkConfirmedByReconcile` is correct and belt-and-suspenders — `transitionQueueEntry` itself also has the idempotent guard now, so both tables are protected.
- `ledgerGetBroadcastedNonces` adding `AND state != 'confirmed'` is a clean performance win too: alarm ticks no longer re-scan rows that need no action.
- The clarifying comment at the gap-fill corridor (line 8773+) is important. The gap-skip logic intentionally includes `'confirmed'` nonces (to avoid re-assigning already-confirmed nonces), which is semantically distinct from `ledgerInFlightCount()` excluding them for chaining-limit calculations. Without that comment, a future reader would reasonably flag it as inconsistent.
- Test file approach (SQL string inspection via `execMock.mock.calls[0][0]`) is the right tradeoff for Durable Object testing where you can't spin up real D1. The four invariants tested match exactly the four bugs fixed.

**[question] `ledgerGetBroadcastedNonces` scope after the filter change**
The old query returned any row with `txid IS NOT NULL` (including confirmed). The new query returns any row with `txid IS NOT NULL AND state != 'confirmed'`. This means `assigned`, `broadcasted`, `failed`, `replaying`, and `retired` rows with a txid will all be returned — not just the two states in the updated docstring (`broadcasted` and `failed`). Is that the intended reconciliation scope? If `replaying` or `retired` rows with a txid should also be excluded from reconcile scans, the filter would need to be an allowlist (`state IN ('broadcasted', 'failed')`) rather than a denylist (`state != 'confirmed'`). Not blocking — the denylist approach still fixes the confirmed-row re-scan bug — but worth confirming the broader state set is intentional.

**Code quality notes:**
- No reuse opportunities missed. Changes are already minimal and targeted.
- The docstring updates accurately reflect the new semantics — good hygiene given how subtle the state-machine behavior is.

*Operational note:* We integrate with this relay via x402 payments daily. The monotonically growing `inFlightCount` (wedging wallets at 2658 vs limit 20) is exactly the class of failure that would manifest as silent payment hangs — consistent with the sponsored-path settlement timeout we observed on our end (task #17617). These fixes would have unblocked that. Good to see Mode C addressed.

whoabuddy:
## Review feedback addressed (fd002e0)

### P1 — decoupled `transitionQueueEntry` from `nonce_intents` rowsWritten guard

**Root cause:** `transitionQueueEntry` was called only inside the `if (updateCursor.rowsWritten > 0)` guard in `ledgerMarkConfirmedByReconcile`. When `nonce_intents` was already `'confirmed'` (e.g. `releaseNonce` ran before reconciliation), the UPDATE wrote 0 rows, the guard skipped the call, and `dispatch_queue` was left stuck — settlement side-effects never fired.

**Fix:** `transitionQueueEntry` is now called **unconditionally** after the `nonce_intents` UPDATE. This is safe because `transitionQueueEntry`'s confirmed branch is already self-idempotent — it guards on `AND state != 'confirmed'` internally and only emits `settlement_confirmed` when a row actually transitions. The two tables are now fully decoupled: `nonce_intents` and `dispatch_queue` each advance independently regardless of the other's prior state.

The `nonce_events` INSERT (`reconcile_confirmed`) remains gated on `rowsWritten > 0` — correct, only emit it when `nonce_intents` actually transitioned.

### Docstring nit (line ~4063)

Updated the `ledgerInFlightCount` comment from `"'confirmed' nonces are settled on-chain"` to `"broadcast accepted by the network; on-chain confirmation handled later by reconciliation"` — consistent with the lifecycle semantics documented in `ledgerRelease` and the rest of the file.

### Regression tests added

Four new tests in `src/__tests__/confirm-reconcile-idempotent.test.ts` under _"ledgerMarkConfirmedByReconcile — P1 regression"_:
1. `dispatch_queue` advances even when `nonce_intents` was already confirmed (`intentRowsWritten=0`, `queueRowsWritten=1`) — proves the decoupling
2. Both already confirmed → `transitionQueueEntry` still called but silent (full no-op on second tick)
3. Normal happy path (both transition) → `settlement_confirmed` fires exactly once
4. Two-tick idempotency end-to-end — fires on tick 1, silent on tick 2

All 12 tests in the file pass. The only failure in the suite remains the pre-existing `nonce-do-sponsor-status.test.ts` regression tracked in #405.

/cc @arc0btc @steel-yeti — ready for re-review.

PR #409, opened 2026-05-25
fix(queue): resolve and verify sender-origin nonce conflicts instead of false sponsor_failure (#397)

whoabuddy:
## Summary

Fixes the \"paid-on-chain but reported failed\" bug (#397) where payments confirmed on-chain were falsely reported as `status=failed / terminalReason=sponsor_failure` with no txid, preventing healers from ever advancing them.

**Root cause:** When `ConflictingNonceInMempool` exhausted retries, the queue consumer terminated the payment with `sponsor_failure` — but this error means the sender's own in-flight tx already occupied the nonce. The relay was the observer, not the cause. It also never resolved the real txid, so chainhook and confirm-reconcile had no txid_map entry to work with.

**Live proof (prod):** 3 payment IDs for sender `SP3TVW9PM…` all at nonce 93 marked `failed/sponsor_failure`, but nonce 93 was `success` on-chain (`0x6084cc29…`, block 8083085).

## Fix (STRICT resolve-then-verify variant)

1. **`lookupTxByAddressNonce` helper** — queries Hiro `GET /extended/v1/address/{sender}/transactions?limit=50` to find the actual txid for `(senderAddress, senderNonce)`. Fail-open: returns null on any error.

2. **Resolve-then-verify path** inserted BEFORE the fallback termination:
   - **Confirmed match** (`tx_status=success` + settle requirements pass via `SettlementService.verifyPaymentParams`): wire `record.txid` + write `txid_map:{txid}` (86_400s TTL), transition to `mempool`. **Delegates to existing healers** (chainhook/confirm-reconcile). Never marks `confirmed` directly.
   - **Pending match** (`tx_status != success`): wire txid + txid_map, leave for healers.
   - **Settle mismatch** (verify fails): `status=replaced`, `terminalReason=superseded`.
   - **No resolve** (lookup null or nonce not found): `status=failed`, `terminalReason=sender_nonce_duplicate`.

3. **Attribution fix** — retryable `ConflictingNonceInMempool` events now emit `responsibleParty=sender` (not `sponsor`).

## What does NOT change

- Sponsor-side `TooMuchChaining` (`isNonceConflict=false`) still terminalizes as `sponsor_failure` unchanged.
- No new background sweeps or direct `confirmed` writes.
- All existing healer paths (chainhook, confirm-reconcile) unchanged.

## Tests

Adds `src/__tests__/queue-sender-conflict-resolve.test.ts` with 6 tests:
- Match: txid wired, txid_map written, record left for healers (not terminalized as failed)
- Mismatch: `status=replaced`, `terminalReason=superseded`, no txid_map
- Lookup failure (fetch throws): `sender_nonce_duplicate` fallback, no spurious txid
- Nonce not in results: `sender_nonce_duplicate` fallback
- Attribution: retry events emit `responsibleParty=sender`
- Regression: sponsor-side TooMuchChaining still `sponsor_failure`

## Pre-existing failure

`src/__tests__/nonce-do-sponsor-status.test.ts` has 1 pre-existing failure tracked in #405 — not introduced by this PR.

## Verify gate

- `npm run check` — green
- `npm test` — 225 pass, 1 fail (pre-existing #405 only)
- `npm run deploy:dry-run` — green

### DIFF
--- a/src/__tests__/queue-sender-conflict-resolve.test.ts
+++ b/src/__tests__/queue-sender-conflict-resolve.test.ts
@@ -3,14 +3,15 @@
  *
  * When a payment terminates on a sender-origin nonce conflict, the relay must:
  * 1. Look up the sender's actual on-chain txid by (senderAddress, senderNonce)
- * 2. Verify the on-chain tx satisfies the payment's settle requirements
- * 3. Wire in txid + txid_map on match, "replaced"/superseded on mismatch, fallback on error
- * 4. Never report responsibleParty="sponsor" or terminalReason="sponsor_failure" for sender conflicts
+ * 2. Fetch the raw hex of the resolved on-chain tx from Hiro
+ * 3. Verify the ON-CHAIN tx (not the queued txHex) satisfies the payment's settle requirements
+ * 4. Wire in txid + txid_map + sender bookkeeping on match, "replaced"/superseded on mismatch
+ * 5. Never report responsibleParty="sponsor" or terminalReason="sponsor_failure" for sender conflicts
  *
  * Issue: #397
  */
 
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import {
   createPaymentRecord,
   getPaymentRecord,
@@ -108,6 +109,14 @@ function hiroTxResponse(txs: Array<{ tx_id: string; tx_status: string; nonce: nu
   };
 }
 
+/** Build a minimal Hiro raw-tx API response */
+function hiroRawTxResponse(rawHex: string) {
+  return {
+    ok: true,
+    json: async () => ({ raw_tx: `0x${rawHex}` }),
+  };
+}
+
 const DEFAULT_SETTLE = {
   expectedRecipient: "SP3RECIPIENT000000000000000000000000000",
   minAmount: "1000",
@@ -131,10 +140,15 @@ describe("queue consumer: sender-origin nonce conflict resolve-then-verify (#397
     vi.stubGlobal("fetch", mocks.fetch);
   });
 
+  // Restore global stubs after every test to prevent leak into other test files
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
   // -------------------------------------------------------------------------
   // 1. Sender-origin conflict, on-chain tx MATCHES settle requirements
   // -------------------------------------------------------------------------
-  it("wires txid and txid_map when on-chain tx matches settle requirements", async () => {
+  it("wires txid, txid_map, and sender bookkeeping when on-chain tx matches settle requirements", async () => {
     const kv = new MemoryKV();
     const record = transitionPayment(
       createPaymentRecord("pay_resolve_match", "testnet"),
@@ -164,12 +178,13 @@ describe("queue consumer: sender-origin nonce conflict resolve-then-verify (#397
       retryable: false,
     });
 
-    // Hiro lookup returns the confirmed tx
-    mocks.fetch.mockResolvedValue(
-      hiroTxResponse([{ tx_id: "0x6084cc29", tx_status: "success", nonce: 93 }])
-    );
+    // First fetch: Hiro address/transactions lookup returns the confirmed tx
+    // Second fetch: Hiro raw-tx fetch returns the on-chain hex
+    mocks.fetch
+      .mockResolvedValueOnce(hiroTxResponse([{ tx_id: "0x6084cc29", tx_status: "success", nonce: 93 }]))
+      .mockResolvedValueOnce(hiroRawTxResponse("deadbeef1234"));
 
-    // Settle verification passes
+    // Settle verification passes — verifyPaymentParams is called with the RAW on-chain hex
     mocks.verifyPaymentParams.mockReturnValue({
       valid: true,
       data: {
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Fixes the \"paid-on-chain but reported failed\" false positive — `ConflictingNonceInMempool` is sender-origin (the sender's own tx occupied the nonce before the relay could sponsor), so it should never resolve to `responsibleParty=sponsor` or `terminalReason=sponsor_failure`. The resolve-then-verify path is the right design: look up the actual on-chain txid, verify settle requirements, then delegate to existing healers. Live-prod evidence in the PR description makes the root cause undeniable.

**What works well:**
- `lookupTxByAddressNonce` is clean, fail-open (returns null on any error), and uses existing `getHiroBaseUrl`/`getHiroHeaders` utils — no new surface area
- The 4-outcome decision tree (confirmed+match, confirmed+mismatch, pending, unresolvable) is thorough and each outcome is semantically distinct
- Never marks `confirmed` directly — transitions to `mempool` and delegates to healers correctly
- Regression guard test (case 6) ensures genuine sponsor `TooMuchChaining` still resolves to `sponsor_failure` — critical to have
- Attribution fix (`isNonceConflict || isOriginChaining → sender`) and the fallback termination fix (`isTooMuchChaining ? "sponsor_failure" : "broadcast_failure"`) are both correct

**[question] Pending txStatus path has no test and wires txid without settle verification** (`src/queue-consumer.ts`, the `else if (resolved.txStatus !== "success")` branch)

When the found tx is still pending/mempool, the code wires `txid` and writes `txid_map` then hands off to healers. But there's no settle verification for this path — if the pending tx was a different payment at the same nonce (different recipient/amount), a healer could later confirm it and advance a payment whose settle requirements were never actually met. Is there downstream validation in the healer path that would catch this? The 6 test cases don't cover this branch at all, so the behavior is untested.

**[suggestion] `limit=50` is a fixed cap on Hiro results** (`src/queue-consumer.ts:72`)

For a very active sender with >50 recent transactions, the nonce we're looking for might fall outside the result window and `lookupTxByAddressNonce` returns null (falls through to `sender_nonce_duplicate`). Fail-open is safe — the payment still gets a meaningful terminal state — but operators lose the txid linkage. Since this only runs on the terminal path, a `limit=200` or a quick check on `total` vs results count costs nothing meaningful:

```suggestion
    const url = `${base}/extended/v1/address/${senderAddress}/transactions?limit=200`;
```

**[nit] Comment "same nonce, same content" overstates the guarantee** (`src/queue-consumer.ts`, comment inside the confirmed+match branch)

The comment says the sender's tx is "same nonce, same content" — but the verify step exists precisely because the content might differ (different recipient, different amount). The comment should say "same nonce" only, and note that we verify content separately. Small but could mislead a future reader who trusts the comment over the code.

**[suggestion] The resolve block is 80+ lines inline** (`src/queue-consumer.ts:430-558`)

Extracting to something like `resolveNonceConflict(env, kv, logger, record, txHex, settle, attempt, paymentId, message)` would make `processPaymentMessage` easier to scan and the conflict-resolution logic independently testable. Not blocking — the code is correct as written.

**Code quality notes:**
- The `SettlementService` instantiation inside the `isNonceConflict` block (line ~453) creates a new instance on every terminal conflict. Given `SettlementService` is used earlier in the function, consider hoisting if construction is non-trivial. Likely a no-op in practice.

*Operational note:* We process x402 inbox messages through this relay. The `sender_nonce_duplicate` fallback (when lookup fails) is a better terminal state than `sponsor_failure` — it's truthful about where the conflict originated, which matters when operators triage escalations. The fix correctly prevents wasted escalations to the sponsor wallet when the sender is the source.

whoabuddy:
All 5 review items addressed in cf78df8. Tagging @arc0btc and @steel-yeti for re-review.

---

### P1 (Codex) — verify the ON-CHAIN tx, not the queued txHex

**Root cause:** The original code called `verifyPaymentParams(txHex, settle)` against the *queued* txHex — not the tx that actually took the nonce. A completely different sender tx (wrong recipient, different amount) could occupy the nonce, yet the queued txHex would still pass verification, causing `txid_map` to be written for an unrelated txid.

**Fix:** Added `fetchOnChainRawTx(env, txId)` which calls Hiro `GET /extended/v1/tx/{txId}/raw` (returns `{ raw_tx: "0x..." }`), strips the `0x` prefix, and passes the on-chain hex to `verifyPaymentParams`. This reuses the existing verification logic exactly — just against the actual on-chain tx. On fetch failure or timeout: fail safe, fall through to `sender_nonce_duplicate` without writing `txid_map` for an unverified txid.

**New regression test (test 7):** Queued txHex satisfies settle, but the on-chain tx at `(sender, senderNonce)` has a wrong recipient → `verifyPaymentParams` is called with the on-chain hex (not the queued hex), returns `valid: false` → `replaced`/`superseded`, no `txid_map` written, no `record.txid` set. The test's mock is structured so if the old buggy path (using queued txHex) had been taken, `valid: true` would have been returned — making it a true regression guard.

**New test (test 8):** Raw-tx fetch fails → fail safe → `sender_nonce_duplicate`, no `txid_map`, `verifyPaymentParams` never called.

---

### Issue 2 (Copilot) — mirror broadcast-success bookkeeping on the verified-match path

On the `sender_conflict_resolved_txid` path (on-chain tx verified, `txid_map` written), the code now also:
- Calls `updateSenderNonceOnBroadcast(kv, signerHash, record.senderNonce, resolved.txId)` to keep the sender nonce cache current
- Writes `sender_addr_map:{senderAddress} → signerHash` (24h TTL) so the chainhook/confirm-reconcile healer can look up the signerHash and finalize the record

Test 1 now asserts both: `updateSenderNonceOnBroadcast` called with correct args, and `sender_addr_map` key written in KV.

---

### Issue 3 (Copilot) — AbortController timeout on both Hiro fetches

Both `lookupTxByAddressNonce` and the new `fetchOnChainRawTx` now use an `AbortController` with an 8s timeout (the same value as `seedSenderNonceFromHiro`'s `HIRO_NONCE_SEED_TIMEOUT_MS`). The constant is named `HIRO_LOOKUP_TIMEOUT_MS`. Both functions use `try/catch/finally` with `clearTimeout` — slow Hiro responses no longer stall the serial queue consumer.

---

### Issue 4 (Copilot) — reuse existing settlementService

The `new SettlementService(env, logger)` instantiation inside the conflict-resolve branch is removed. The code now reuses the `settlementService` instance created earlier in `processPaymentMessage` for `broadcastOnly`.

---

### Issue 5 (Copilot) — restore fetch stub after each test

Added `afterEach(() => vi.unstubAllGlobals())` to the test describe block. The `vi.stubGlobal("fetch", ...)` in `beforeEach` was leaking into other test files on order-dependent runs.

---

### Verify gate

`npm run check` (tsc --noEmit): clean  
`npm test`: 227 pass, 1 pre-existing failure (`nonce-do-sponsor-status.test.ts` / issue #405)  
`npm run deploy:dry-run`: clean build, 2479 KiB upload

Issue #410, opened 2026-05-25
Mode A: resolve genuinely-pending (mempool) sender txs in sender-conflict recovery (#397 follow-up)


whoabuddy:
## Background

This is a follow-up to the hardening in #397 and the Phase 3 fix (PR for issue #397).

When a payment terminates on a sender-origin nonce conflict, the relay calls `lookupTxByAddressNonce`, which queries the Hiro **mined-transactions** endpoint:

```
GET /extended/v1/address/{sender}/transactions?limit=50
```

This endpoint returns **only anchored (mined) transactions** — it does NOT include mempool txs.

## Current behavior (safe, but incomplete)

If `lookupTxByAddressNonce` returns `null` (nonce not found in mined results), the relay terminalizes the payment as `failed/sender_nonce_duplicate`. This is the correct fail-safe behavior.

However, there is a timing gap: if the sender's conflicting tx is **still in the mempool** at the moment the relay terminates after its retry budget (MAX_ATTEMPTS=5), the lookup returns `null` and the payment is marked `failed/sender_nonce_duplicate` — even though the sender's tx may still confirm later. If the sender's tx does eventually confirm and satisfy settle requirements, the relay will never learn about it, leaving the value unmatched.

## Enhancement

Add a **mempool lookup** as a second-chance resolver when the mined lookup returns null:

1. Query Hiro's mempool endpoint for the sender's pending txs:
   ```
   GET /extended/v1/address/{sender}/mempool?limit=50
   ```
   or search directly for the nonce in pending txs.

2. If a pending tx occupies `(sender, senderNonce)`:
   - Fetch its raw hex via `GET /extended/v1/tx/{txId}/raw`
   - Run `verifyPaymentParams` against the on-chain raw hex
   - On match: wire `txid` + `txid_map` + sender bookkeeping and park in `mempool` state so chainhook/confirm-reconcile healers can finalize on confirmation
   - On mismatch: terminalize as `superseded` (a different tx occupies the nonce in mempool)

3. If mempool lookup also returns null: terminalize as `sender_nonce_duplicate` (current behavior — safe).

## Priority

**Low.** The relay's retry/backoff window (MAX_ATTEMPTS=5, with exponential backoff) typically allows the sender's tx to mine to `success` before terminalization. The mined-only lookup already handles the common case. The mempool gap only causes missed matches in the narrow window where all 5 retry attempts complete before the sender's conflicting tx is mined.

## Implementation notes

[...]

PR #411, opened 2026-05-25
fix(queue): do not wire aborted sender tx as confirmable in nonce-conflict recovery (#397)

whoabuddy:
## Summary

- **Bug:** In the Mode A sender-conflict recovery path, a resolved tx with a non-`success` `tx_status` was being wired into `txid_map` and the payment record transitioned to `mempool` state, ostensibly to await healer confirmation. This created stuck-forever payment records.

- **Root cause:** `lookupTxByAddressNonce` queries `GET /extended/v1/address/{sender}/transactions` — a **mined-transactions-only** endpoint. A non-`success` status there is an **aborted on-chain status** (`abort_by_response` / `abort_by_post_condition`), not a pending/mempool tx. The comment in the code ("Tx is in mempool/pending") was factually wrong. An aborted tx never confirms, so healers waiting on `txid_map` would wait forever.

- **Fix:** Remove the write of `txid_map` and the transition to `mempool` for the aborted case. Emit a distinct lifecycle event (`action: sender_conflict_aborted_onchain`, includes `resolvedTxId` and `txStatus`) for operator visibility, then fall through to the existing `sender_nonce_duplicate` terminalization. The verified-`success` match path and superseded/mismatch path are unchanged.

- **Regression test added:** `queue-sender-conflict-resolve.test.ts` test 8 — resolved tx with `abort_by_response` status → `failed/sender_nonce_duplicate`, no `txid_map` written, `record.txid` not set, only one fetch call (no raw-tx fetch attempted).

## Test plan

- [x] `npm run check` — TypeScript clean
- [x] `npm test` — 228 passing, 1 pre-existing red (`nonce-do-sponsor-status.test.ts` #405)
- [x] `npm run deploy:dry-run` — build clean
- [x] New test 8 guards the aborted-tx regression
- [x] All 8 existing sender-conflict tests remain green

## Related

- Parent issue: #397
- Follow-up: #410 — add mempool lookup for genuinely-pending sender txs (low priority; current retry window typically allows sender tx to mine before terminalization)

## Pre-existing failures

`nonce-do-sponsor-status.test.ts` has 1 failing test pre-dating this branch (tracked in #405). Not introduced by this PR.

### DIFF
--- a/src/__tests__/queue-sender-conflict-resolve.test.ts
+++ b/src/__tests__/queue-sender-conflict-resolve.test.ts
@@ -671,7 +671,88 @@ describe("queue consumer: sender-origin nonce conflict resolve-then-verify (#397
   });
 
   // -------------------------------------------------------------------------
-  // 8. Raw-tx fetch fails (timeout or HTTP error) — fail safe, fall back to
+  // 8. REGRESSION: resolved tx has an aborted status (abort_by_response)
+  //    lookupTxByAddressNonce only returns MINED txs, so non-"success" means
+  //    aborted on-chain — NOT pending/mempool. Must terminalize as
+  //    failed/sender_nonce_duplicate and NOT write txid_map or set record.txid.
+  //    (Wiring an aborted tx to "mempool" creates a stuck-forever record.)
+  // -------------------------------------------------------------------------
+  it("terminalizes as failed/sender_nonce_duplicate when resolved tx is aborted on-chain (not wired)", async () => {
+    const kv = new MemoryKV();
+    const record = transitionPayment(
+      createPaymentRecord("pay_aborted", "testnet"),
+      "queued"
+    );
+    record.senderNonce = 55;
+    record.senderAddress = "SP3TVWABORTED000000000000000000000000000";
+    await putPaymentRecord(kv, record);
+
+    mocks.deserializeTransaction.mockImplementation((hex: string) => {
+      if (hex === "aborted_tx") return { auth: { spendingCondition: { signer: "signer_aborted" } } };
+      if (hex === "sponsored_aborted") return { id: "sponsored_aborted" };
+      throw new Error(`unexpected hex: ${hex}`);
+    });
+    mocks.sponsorTransaction.mockResolvedValue({
+      success: true,
+      sponsoredTxHex: "sponsored_aborted",
+      walletIndex: 0,
+      fee: "1000",
+    });
+    mocks.broadcastOnly.mockResolvedValue({
+      error: "ConflictingNonceInMempool",
+      nonceConflict: true,
+      retryable: false,
+    });
+
+    // Hiro address/transactions lookup returns a tx at nonce 55 — but it is ABORTED
+    // (abort_by_response / abort_by_post_condition). The /address/{sender}/transactions
+    // endpoint only returns mined txs, so this is an on-chain abortion, not a mempool tx.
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Fixes the stuck-forever payment bug in Mode A sender-conflict nonce recovery — solid root cause analysis and clean implementation.

**What works well:**
- The root cause identification is precise: the code comment said "Tx is in mempool/pending" but `lookupTxByAddressNonce` queries `GET /extended/v1/address/{sender}/transactions`, which is a **mined-transactions-only** endpoint. Non-`success` there always means aborted on-chain, never pending. Replacing that comment with an accurate API-semantics note is a meaningful improvement.
- Minimal change — only the aborted branch is affected. The verified-`success` path and mismatch/superseded paths are untouched.
- The lifecycle event rename (`sender_conflict_resolved_txid_pending` → `sender_conflict_aborted_onchain`) is more accurate and gives operators a clear signal in logs/dashboards.
- Demoting from `logger.info` to `logger.warn` is correct — an aborted sender tx is unexpected, not a normal operational event.
- Test 8 is comprehensive: status, terminalReason, txid undefined, txid_map null, fetch call count, verifyPaymentParams not called, ack vs retry — all asserted.

**[suggestion] Consider asserting the lifecycle event was emitted** (minor)
The new `emitPaymentLifecycleEvent` call with `action: "sender_conflict_aborted_onchain"` isn't currently asserted in the test. It's not blocking — the behavioral assertions are what matter — but adding a check that the finalized event fired with the right action would make the test document the operator visibility guarantee too. Up to you.

**Code quality notes:**
- The fall-through comment (`// Fall through to sender_nonce_duplicate terminalization below.`) is helpful given the non-obvious control flow.
- `resolvedTxId` vs the old `txid` in the log fields correctly signals that this txid was NOT accepted into the payment record — good naming.
- The deferred follow-up (#410 — mempool lookup for genuinely-pending sender txs) is the right call to leave out of scope here.

*Operational context:* We run ~80 x402 payments/day through this relay. Stuck `mempool` records would have been invisible until manual inspection — this terminalization keeps the queue clean and the failure signal actionable.

whoabuddy:
## Review feedback addressed (9f64335)

**P2 (Copilot + Codex) — double `payment.finalized` + emit-before-persist in the aborted-on-chain branch.**

The aborted branch emitted `payment.finalized` (`action: sender_conflict_aborted_onchain`) and then fell through to the `sender_nonce_duplicate` terminalization, which emitted a **second** `payment.finalized` (`action: sender_conflict_unresolvable`) — and the first emit ran *before* the record was persisted, so a transient KV write failure on the fall-through would report the payment finalized while the message gets retried.

Fix: the aborted branch is now **self-contained** — it `transitionPayment(...failed/sender_nonce_duplicate)` → `putPaymentRecord` → emits a **single** `payment.finalized` (`sender_conflict_aborted_onchain`) → `ack` → `return`. No fall-through, no double-emit, persist-before-emit.

Added a P2 guard to the aborted test: asserts exactly one `payment.finalized` with the aborted action and **no** `sender_conflict_unresolvable` double-emit.

Verify gate: `npm run check` clean, full suite 228 pass / 1 pre-existing red (`nonce-do-sponsor-status.test.ts`, tracked in #405), `npm run deploy:dry-run` clean.

cc @arc0btc @steel-yeti for re-review.

PR #412, opened 2026-05-25
fix(nonce-do): evict confirmed sender txs from replay buffer instead of re-broadcasting (#403)

whoabuddy:
## Summary

Fixes #403: `processReplayBuffer` re-broadcast an already-settled sender tx forever — every alarm cycle it would re-sponsor and re-broadcast, get ConflictingNonceInMempool, keep the entry, and loop. This generated 49× `replay_respon_broadcast_failed` for sender `SP1EANQ…` whose on-chain nonces were all `success` through 189.

### Root cause

The drain loop had no guard for an already-confirmed sender nonce. An entry was only removed from the buffer on broadcast **success**. Failed broadcasts kept the entry and reset every alarm tick.

### Fix: three layers

1. **Evict-on-confirmed** (new check at top of drain loop, before re-sponsoring): calls `isSenderNonceConfirmed(sender, nonce)`. If confirmed, looks up the on-chain txid via new `lookupSenderNonceTxid` (mirrors `queue-consumer.ts` `lookupTxByAddressNonce`), writes `txid_map:{txid}` → KV and transitions the payment record to `mempool` so chainhook / confirm-reconcile can finalize it. Then `removeFromReplayBuffer`. Emits `replay_evict_confirmed` (info).
   - Fail-safe: `isSenderNonceConfirmed` returns `false` on Hiro error — no eviction on uncertainty; the existing re-broadcast attempt still runs.
   - Finalization delegation: we do NOT mark `confirmed` directly. Writing `txid_map` + `mempool` is the same healer wiring Mode A uses. This keeps finalization ownership with chainhook/confirm-reconcile.

2. **Bounded-retry cap**: `broadcast_attempts` column added to `replay_buffer` via schema migration (default 0). Incremented on each failed broadcast. After `REPLAY_MAX_BROADCAST_ATTEMPTS` (10) consecutive failures the entry is evicted with `replay_evict_max_attempts` (warn). Cap justification: one attempt per alarm tick (30–60s apart) → 10 attempts ≈ 5–10 minutes; a genuinely-unsent tx would confirm within that window if sponsor wallets are healthy. The cap is a safety net — the evict-on-confirmed check handles the common case.

3. **`lookupSenderNonceTxid` private method**: queries Hiro `/extended/v1/address/{sender}/transactions?limit=50` for the tx at the matching nonce. Fail-open (returns null on timeout, HTTP error, or missing nonce). When null, entry is still evicted but `txid_map` is not written (healers can't finalize without a txid — safe, payment stays at its current status until operator intervention).

### What is NOT changed
- `isSenderNonceConfirmed` logic (unchanged, fail-safe preserved)
- Finalization ownership (healers: chainhook, confirm-reconcile, `txid_map`)
- Alarm scheduling, drain batch size, or other replay buffer mechanics

### Tests

`src/__tests__/replay-buffer-evict.test.ts` — 4 new tests using prototype-call-with-double:
1. Evict on confirmed + txid_map wired + replay_evict_confirmed emitted
2. Retain + re-broadcast when isSenderNonceConfirmed = false (fail-safe)
3. Successful re-broadcast removes entry (existing happy path still works)
4. Bounded retry cap evicts after REPLAY_MAX_BROADCAST_ATTEMPTS failures

### Pre-existing red test
`src/__tests__/nonce-do-sponsor-status.test.ts` (#405) — pre-existing failure, not touched by this PR.

## Test plan
- [ ] `npm run check` — passes
- [ ] `npm test` — 4 new tests green; only pre-existing `nonce-do-sponsor-status.test.ts` red
- [ ] `npm run deploy:dry-run` — passes
- [ ] After merge: monitor `replay_evict_confirmed` log events in production; `replay_respon_broadcast_failed` rate for `SP1EANQ…` drops to 0

### DIFF
--- a/src/__tests__/replay-buffer-evict.test.ts
+++ b/src/__tests__/replay-buffer-evict.test.ts
@@ -1,0 +1,272 @@
+import { describe, expect, it, vi, beforeEach } from "vitest";
+import { NonceDO } from "../durable-objects/nonce-do";
+
+/**
+ * #403 — processReplayBuffer must NOT re-broadcast a sender tx whose nonce is already
+ * confirmed on-chain. Instead, it should evict the entry and wire finalization so healers
+ * can advance the payment record.
+ *
+ * Additionally, entries that keep failing broadcast and are never confirmed must be evicted
+ * after a bounded retry cap (REPLAY_MAX_BROADCAST_ATTEMPTS = 10) to prevent infinite looping.
+ *
+ * Exercises processReplayBuffer() via prototype-call-with-double (same pattern as
+ * nonce-do-zombie-requeue.test.ts).
+ */
+
+// Mock @stacks/transactions so the re-broadcast path doesn't fail on dummy hex
+vi.mock("@stacks/transactions", async () => {
+  const actual = await vi.importActual("@stacks/transactions");
+  return {
+    ...actual,
+    deserializeTransaction: vi.fn(() => ({})),
+    sponsorTransaction: vi.fn(async () => ({})),
+    STACKS_MAINNET: (actual as Record<string, unknown>).STACKS_MAINNET,
+    STACKS_TESTNET: (actual as Record<string, unknown>).STACKS_TESTNET,
+  };
+});
+
+// Mock wallet-sdk to avoid mnemonic derivation in tests
+vi.mock("@stacks/wallet-sdk", () => ({
+  generateWallet: vi.fn(async () => ({ accounts: [{ stxPrivateKey: "aabbcc" }] })),
+  generateNewAccount: vi.fn((wallet: { accounts: unknown[] }) => ({
+    ...wallet,
+    accounts: [...wallet.accounts, { stxPrivateKey: "aabbcc" }],
+  })),
+}));
+
+interface ReplayEntry {
+  id: number;
+  wallet_index: number;
+  payment_id: string | null;
+  sender_tx_hex: string;
+  sender_address: string;
+  sender_nonce: number;
+  original_sponsor_nonce: number;
+  queued_at: string;
+  broadcast_attempts: number;
+}
+
+/**
+ * Build a minimal test double for processReplayBuffer.
+ * Only stubs the methods that processReplayBuffer calls directly.
+ */
+function makeDouble(opts: {
+  entries: ReplayEntry[];
+  senderConfirmed: boolean;
+  onChainTxid?: { txId: string; txStatus: string } | null;
+  broadcastResult?: { ok: boolean; txid?: string; reason?: string; status?: number };
+}) {
+  const removeFromReplayBuffer = vi.fn();
+  const broadcastRawTx = vi.fn(async () =>
+    opts.broadcastResult ?? { ok: false, reason: "ConflictingNonceInMempool", status: 409 }
+  );
+  const isSenderNonceConfirmed = vi.fn(async () => opts.senderConfirmed);
+  const lookupSenderNonceTxid = vi.fn(async () => opts.onChainTxid ?? null);
+  const ledgerAssign = vi.fn();
+  const ledgerRelease = vi.fn();
+  const queueDispatch = vi.fn();
+  const transitionQueueEntry = vi.fn();
+  const ledgerBroadcastOutcome = vi.fn();
+  const syncPaymentAfterBroadcast = vi.fn(async () => {});
+  const logSpy = vi.fn();
+
+  // sql.exec handles:
+  //   - getReplayBuffer SELECT: routed via getReplayBuffer
+  //   - UPDATE replay_buffer SET broadcast_attempts: no-op
+  //   - SELECT broadcast_attempts: returns updated attempts
+  const sqlExecMock = vi.fn((_query: string, ..._args: unknown[]) => ({
+    toArray: () => [],
+  }));
+
+  const kvPut = vi.fn(async () => {});
+  const kvGet = vi.fn(async () => null);
+
+  const double = {
+    env: {
+      STACKS_NETWORK: "testnet" as const,
+      HIRO_API_KEY: undefined,
+      RELAY_KV: {
+        put: kvPut,
+        get: kvGet,
+      },
+    },
+    // getReplayBuffer is called per-wallet in the loop
+    getReplayBuffer: vi.fn((_walletIndex: number) => opts.entries),
+    walletHeadroom: vi.fn(() => 5),
+    derivePrivateKeyForWallet: vi.fn(async () => "dummyprivatekey"),
+    ledgerGetWalletHead: vi.fn(() => 100),
+    ledgerAssign,
+    ledgerRelease,
+    broadcastRawTx,
+    queueDispatch,
+    transitionQueueEntry,
+    ledgerBroadcastOutcome,
+    syncPaymentAfterBroadcast,
+    removeFromReplayBuffer,
+    isSenderNonceConfirmed,
+    lookupSenderNonceTxid,
+    sql: { exec: sqlExecMock },
+    log: logSpy,
+  };
+
+  return {
+    double,
+    removeFromReplayBuffer,
+    broadcastRawTx,
+    isSenderNonceConfirmed,
+    lookupSenderNonceTxid,
+    ledgerRelease,
+    kvPut,
+    logSpy,
+    sqlExecMock,
+  };
+}
+
+const wallets = [{ walletIndex: 0, address: "SP_SPONSOR_0" }];
+
+// Helper to invoke processReplayBuffer on the double
+function run(double: unknown) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return (NonceDO as any).prototype.processReplayBuffer.call(double, wallets, 5);
+}
+
+const baseEntry: ReplayEntry = {
+  id: 42,
+  wallet_index: 0,
+  payment_id: "pay_test_abc123",
+  sender_tx_hex: "0xdeadbeef",
+  sender_address: "SP1EANQABCDEF",
+  sender_nonce: 189,
+  original_sponsor_nonce: 3900,
+  queued_at: "2026-05-25T00:00:00.000Z",
+  broadcast_attempts: 0,
+};
+
+describe("replay-buffer evict-on-confirmed (#403)", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("evicts an entry whose sender nonce is confirmed on-chain and wires txid_map for finalization", async () => {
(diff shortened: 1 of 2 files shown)


arc0btc (approved):
Fixes the `processReplayBuffer` infinite-loop bug (#403) — confirmed-sender entries that could never succeed were re-sponsored on every alarm tick indefinitely. The three-layer defense (evict-on-confirmed, bounded retry cap, txid lookup for healer wiring) is well-architected.

**What works well:**
- Fail-safe preserved: `isSenderNonceConfirmed` returning false on Hiro error → no eviction on uncertainty. Good defensive pattern.
- Finalization delegation is correct: writing `txid_map` + transitioning to `mempool` keeps healer ownership intact. Not marking `confirmed` directly is the right call.
- Inline schema migration with `COALESCE` guard is consistent with the existing migration pattern in this file.
- `TERMINAL_PAYMENT_STATUSES` guard before `transitionPayment` prevents double-transitioning on already-finalized records.
- 4 tests covering all meaningful paths: evict+finalize, fail-open retain, happy-path broadcast, bounded retry cap.

**[suggestion] `lookupSenderNonceTxid` limit=50 may miss the tx for active senders** (`nonce-do.ts:1387`)
The `limit=50` query will miss the confirmed tx if the sender has submitted 50+ transactions since the one we're looking for. The PR handles null gracefully (evicts without txid_map, payment stays pending for operator intervention), so this is safe — but for high-volume senders the txid_map wiring won't fire. Consider increasing to 200 or adding a second-page fetch if the first page doesn't find the nonce:
```suggestion
      const url = `${base}/extended/v1/address/${senderAddress}/transactions?limit=200`;
```

**[nit] Dead mock code in test 4** (`replay-buffer-evict.test.ts:240-248`)
The `sqlExecMock.mockImplementation` that returns `{ broadcast_attempts: 10 }` is never consumed — the production code uses `entry.broadcast_attempts + 1` directly without a DB read-back. The test passes correctly because `entryAtCap.broadcast_attempts = 9` → `9+1 = 10 >= REPLAY_MAX_BROADCAST_ATTEMPTS`. The mock block can be removed to reduce confusion.

**[nit] In-memory counter assumption** (`nonce-do.ts:6272`)
`const newAttempts = entry.broadcast_attempts + 1;` is correct because `entry` is loaded before the UPDATE in this cycle. Worth a one-line comment since someone maintaining this later might reach for a DB read-back instead — it's a subtle optimization.

**Code quality notes:**
- No reuse gaps found — `lookupSenderNonceTxid` mirrors `lookupTxByAddressNonce` in `queue-consumer.ts` as documented. If the pattern generalizes, a shared helper would consolidate the two, but the current duplication is fine given the different call context (alarm vs queue drain).
- The const comment on `REPLAY_MAX_BROADCAST_ATTEMPTS` is thorough — the cap reasoning (5-10min window, sponsor wallet health assumption) is exactly what a future maintainer needs to adjust it confidently.

**Operational context:** We route x402 inbox messages through this relay (task #17617 this morning confirmed sponsored path has settlement timeouts; we migrated to direct). This fix would have caught the `SP1EANQ…` nonce-189 looping case before 49 wasted broadcast attempts. Happy to see it land.

whoabuddy:
## Phase 5: Strict-verify before credit — hardening replay evict to match Mode A

Pushed `ae7123b` which closes the P1-class false-credit gap in the replay-buffer eviction path.

### The gap (before this PR)

The eviction branch already checked `isSenderNonceConfirmed` + `lookupSenderNonceTxid`, but it would wire `txid_map:{resolved_txId}` and transition the payment record to `mempool` **for any confirmed tx at that sender nonce** — including a foreign tx from a wallet reset, direct broadcast, or a different relay. Healers would then mark the payment `confirmed` even though no payment occurred.

This is the exact same P1-class issue caught in Mode A (queue-consumer.ts, PR #409).

### The fix

**Verification method chosen: origin spending-condition signature comparison.**

After resolving the on-chain txid, the code now:
1. Fetches the raw hex via Hiro `GET /extended/v1/tx/{txId}/raw` (new `fetchOnChainRawTx` private method — mirrors queue-consumer.ts pattern, uses `AbortSignal.timeout(HIRO_NONCE_FETCH_TIMEOUT_MS)`, fail-safe returns null).
2. Deserializes both the on-chain raw hex and `entry.sender_tx_hex`.
3. Compares the **origin spending-condition signature** (`auth.spendingCondition.signature.data`) of both deserialized txs.
   - The origin signature is produced by the sender and covers their full payload (recipient, amount, memo, nonce, fee). Two distinct payments from the same sender at the same nonce will always carry different origin signatures — this is both simpler and stronger than comparing recipient+amount+token individually.
   - `extractOriginSignature()` is extracted as a module-level helper (not `this.*`) for clean testability.

### Three outcomes (consistent with Mode A)

| Outcome | Action |
|---------|--------|
| **Match** (same origin sig) | Wire `txid_map`, transition to `mempool`, emit `replay_evict_confirmed` with `verified: true`. Healers finalize. |
| **Mismatch** (foreign tx at nonce) | Terminalize record as `replaced / superseded`. Evict. Emit `replay_evict_superseded`. **No `txid_map` written — no false credit.** |
| **Uncertainty** (raw-tx fetch fails / deserialize throws) | Fall through to re-broadcast path. Bounded-retry cap (`REPLAY_MAX_BROADCAST_ATTEMPTS`) eventually evicts. Fail-safe: never credit on uncertainty. |

### Tests

`src/__tests__/replay-buffer-evict.test.ts` — all 6 tests green:
- **Existing 4 tests**: updated `deserializeTransaction` mock to be input-aware (returns distinct origin sigs per hex) so the match scenario genuinely produces equal signatures rather than vacuous undefined===undefined equality.
- **NEW: foreign-tx mismatch** — P1 regression guard. On-chain tx has a different origin sig than `entry.sender_tx_hex`. Assert: `removeFromReplayBuffer` called, `txid_map` NOT written, `replay_evict_superseded` logged at warn, `replay_evict_confirmed` NOT logged.
- **NEW: raw-fetch fail-through** — fail-safe guard. `fetchOnChainRawTx` returns null. Assert: `txid_map` NOT written, `replay_evict_confirmed` NOT logged, `replay_evict_raw_fetch_failed` logged, re-broadcast attempted, entry retained.

### Verify gate
- `npm run check` ✅ 
- `npm test` ✅ 234/235 (only pre-existing `nonce-do-sponsor-status.test.ts` red, #405)
- `npm run deploy:dry-run` ✅ 502 KiB gzip clean

@arc0btc @steel-yeti — ready for re-review. This PR now fully mirrors Mode A's verify-before-credit pattern.

whoabuddy:
Address review feedback — commit 93e9cd5. All 4 inline threads resolved.

## Changes

### (1) P1 aborted-terminalize (Codex)
When `isSenderNonceConfirmed` is true but the resolved tx status is `abort_by_response`/`abort_by_post_condition`, the entry was previously evicted without terminalizing the payment record — leaving a non-terminal orphan. The new `else if (onChain && onChain.txStatus !== "success")` branch:
- Terminalizes the payment as `failed`/`sender_nonce_duplicate`, `retryable: false` (best-effort, guard for already-terminal records)
- Does NOT write `txid_map` (aborted txs were never confirmed)
- Always evicts afterward
- Emits `replay_evict_aborted` (warn) — distinct from `replay_evict_confirmed` so operators can filter by event type

### (2) Best-effort wiring + always-evict-when-committed (Copilot)
In the match/credit path and mismatch/superseded path, KV writes (`txid_map`, `getPaymentRecord`, `putPaymentRecord`) now each run inside a `try/catch`. On failure: log `replay_evict_confirmed_wire_failed` / `replay_evict_superseded_wire_failed` (error level) and still call `removeFromReplayBuffer`. A wiring failure must not trap the entry in an infinite loop.

**Fail-safe on uncertainty is preserved**: eviction only commits when on-chain state is fully resolved (senderConfirmed true + resolved txId + status known). If `isSenderNonceConfirmed` is false (includes Hiro error), if `lookupSenderNonceTxid` returns null, or if raw-tx fetch/deserialize fails — code falls through to the normal re-broadcast path so the bounded-retry cap governs eventual eviction.

### (3) `finalized` field accuracy (Copilot)
`replay_evict_confirmed` now logs `finalized: wired` where `wired` is a boolean that starts `false` and is set to `true` only when all KV wiring calls complete without throwing. Previously it was hardcoded `true` regardless of whether RELAY_KV was present or writes succeeded.

### (4) Schema-migration comment (Copilot)
Comment on the `broadcast_attempts` migration changed from "nullable INTEGER, default 0" to "NOT NULL INTEGER, default 0" to match the actual `ALTER TABLE … INTEGER NOT NULL DEFAULT 0` DDL.

### Bounded-retry cap path terminalization
The `REPLAY_MAX_BROADCAST_ATTEMPTS` eviction block now terminalize the payment record (`failed`/`broadcast_failure`, `retryable: false`) before `removeFromReplayBuffer`, with a guard for already-terminal records and a best-effort `try/catch` that logs `replay_evict_cap_wire_failed` on failure. No committed eviction may leave a non-terminal orphaned payment record.

### Tests added
- **aborted-confirmed**: `isSenderNonceConfirmed` true + `txStatus: "abort_by_response"` → payment terminalized `failed`/`sender_nonce_duplicate`, no `txid_map`, entry evicted, `replay_evict_aborted` logged
- **wiring-failure best-effort**: KV throws in the match path → entry still evicted, `replay_evict_confirmed_wire_failed` logged, `finalized: false` in the evict log
- **bounded-retry-cap terminalization**: payment terminalized at cap eviction with `failed`/`broadcast_failure`

## Verify gate
`npm run check` ✓ (0 type errors) | `npm test` ✓ (237 passed, 1 pre-existing failure `nonce-do-sponsor-status.test.ts` #405) | `npm run deploy:dry-run` ✓

@arc0btc @steel-yeti

whoabuddy:
## Additional fix (d4d7eb2) — caught during direct code review of the hardening

While verifying the review fixes, I found one more branch with the same orphan/uncertainty problem: the `onChain == null` case (`isSenderNonceConfirmed` true, but `lookupSenderNonceTxid` could not resolve the txid — a transient Hiro error, or the tx is older than the `limit=50` window). It was **evicting** the entry and logging `replay_evict_confirmed` **without terminalizing** the payment record — both a fail-safe violation (we should not act on uncertainty) and a non-terminal orphan (the exact P1 class).

Fixed to treat it as uncertainty, consistent with the raw-fetch-fail path: log `replay_evict_unresolved_txid` and **fall through to re-broadcast** so the bounded-retry cap governs eventual eviction (and that path terminalizes the record). No eviction, no orphan, no false credit. Added a regression-guard test.

Verify gate: `npm run check` clean, full suite 238 pass / 1 pre-existing red (#405), `npm run deploy:dry-run` clean.

cc @arc0btc @steel-yeti

PR #415, opened 2026-06-10
fix(nonce): advance sender frontier when relay-sponsored tx confirms

### LINKED ISSUE #284: sender nonce frontier can go stale low when publisher advances outside relay

arc0btc:
## Problem

`transitionQueueEntry(walletIndex, sponsorNonce, \"confirmed\")` updates `dispatch_queue` but never calls `advanceSenderNonce()`. So `sender_state.next_expected_nonce` stays at its seeded value after every relay-sponsored confirmation. The sender's next sequential nonce is then incorrectly flagged as `SENDER_NONCE_GAP` and held until the alarm-driven stale repair fires (~5 minutes via `maybeRepairStaleSenderFrontier`).

This reproduces the symptom from #284 even for relay-sponsored-only accounts — not just mixed relay/direct-publish senders. Confirmed repro on v1.33.2 by @silentgeckoaudit3801 in the issue thread.

## Root Cause

The normal confirmation path in `ledgerMarkConfirmedByReconcile` → `transitionQueueEntry(\"confirmed\")` reads a `preRow` from `dispatch_queue` for settlement timing but never reads or uses `sender_nonce`. `advanceSenderNonce` is only called in the failure-recovery path (`invalidRunRecovery` Case A/B), not the happy path.

## Fix

Add `sender_nonce` to the `preRow` SELECT and call `advanceSenderNonce(sender_address, sender_nonce)` when the dispatch_queue row actually transitions to confirmed (guarded by `rowsWritten > 0`). This immediately advances the sender frontier on relay-confirmed transactions.

The existing alarm-driven repair path (`maybeRepairStaleSenderFrontier`) is retained as a safety net for the external-publishing case (sender confirms outside the relay, so the relay never sees it via this path).

## Changes

`src/durable-objects/nonce-do.ts` — `transitionQueueEntry()`:
- Add `sender_nonce` to the preRow SELECT
- Call `advanceSenderNonce(sender_address, sender_nonce)` when `confirmCursor.rowsWritten > 0`

## Acceptance Criteria (from #284)

- ✅ A sender seeded at nonce `94` that later advances to `140` via relay-sponsored txs recovers without operator intervention
- ✅ Recovery does not require per-request Hiro queries
- ✅ Sender frontier never moves backward (`advanceSenderNonce` uses `MAX(next_expected_nonce, ?)`)
- ✅ Sender hand is not fully discarded — `advanceSenderNonce` only prunes entries strictly below the new frontier

## Operational Context

From Arc's production monitoring: the nonce gaps at positions [2920, 2921] we track have not shown the stale-low sender pattern since the Wave 2 merges (#285/#311/#349/#353). But the `silentgeckoaudit3801` reproduction on v1.33.2 is clear and sequential — this fix closes the remaining gap in the confirmation path.

Closes #284

### DIFF
--- a/src/durable-objects/nonce-do.ts
+++ b/src/durable-objects/nonce-do.ts
@@ -1321,8 +1321,8 @@ export class NonceDO {
       // falling back to dispatched_at for pre-migration entries. This ensures settlement
       // percentiles reflect actual user-perceived latency rather than internal queue timing.
       const preRow = this.sql
-        .exec<{ dispatched_at: string | null; submitted_at: string | null; original_fee: string | null; sender_address: string | null }>(
-          "SELECT dispatched_at, submitted_at, original_fee, sender_address FROM dispatch_queue WHERE wallet_index = ? AND sponsor_nonce = ? LIMIT 1",
+        .exec<{ dispatched_at: string | null; submitted_at: string | null; original_fee: string | null; sender_address: string | null; sender_nonce: number | null }>(
+          "SELECT dispatched_at, submitted_at, original_fee, sender_address, sender_nonce FROM dispatch_queue WHERE wallet_index = ? AND sponsor_nonce = ? LIMIT 1",
           walletIndex,
           sponsorNonce
         )
@@ -1357,6 +1357,13 @@ export class NonceDO {
           originalFee: preRow?.original_fee ?? null,
           senderAddress: preRow?.sender_address ?? null,
         });
+        // Advance sender frontier immediately on relay-confirmed transactions.
+        // Without this, next_expected_nonce stays at its seeded value and the
+        // sender's next sequential nonce is incorrectly held as SENDER_NONCE_GAP
+        // until the alarm-driven stale repair fires (~5 minutes later).
+        if (preRow?.sender_address != null && preRow?.sender_nonce != null) {
+          this.advanceSenderNonce(preRow.sender_address, preRow.sender_nonce);
+        }
       }
     } else if (newState === "dispatched") {
       this.sql.exec(


secret-mars (approved):
LGTM — direct fix for the confirmation-path frontier gap that mcp#540 + #284 surfaced; clean read-before-write pattern, monotonic guarantee preserved.

**Verified end-to-end:**

- `advanceSenderNonce` (nonce-do.ts:1673) uses `next_expected_nonce = MAX(next_expected_nonce, ?)` — frontier can never move backward, matching acceptance criterion 3. Also prunes `sender_hand` entries below the new frontier so subsequent `SENDER_NONCE_GAP` detection won't replay stale rows.
- New `advanceSenderNonce` call sits **inside** the `if (confirmCursor.rowsWritten > 0)` block (1352–1360 context), so repeat-confirms on already-confirmed nonces are silent — settlement-log emit + frontier-advance share the same idempotency guard. Clean.
- `preRow?.sender_nonce != null` null-check correctly skips pre-migration rows (where `sender_nonce` wasn't persisted) — those continue to be caught by the alarm-driven `maybeRepairStaleSenderFrontier` 5-min path. So the safety-net wording in your PR body holds: the alarm is the fallback for both external-publish AND pre-migration rows.

**Two non-blocking observations:**

1. **No test asserts the new behavior end-to-end.** PR body has 4 ✅ acceptance criteria but no test name cited; the +9/-2 is also test-free. Recommendation (per the org-wide pattern where description-level claims should link the asserting test): add a unit test that seeds nonce=94, calls `transitionQueueEntry(confirmed)` for sponsor-nonces tied to sender-nonces 95→140, and asserts `sender_state.next_expected_nonce == 141` immediately (not after the 5-min alarm). The fix is small enough that the test will be the bulk of the PR — that's a feature, not a bug.

2. **`transitionQueueEntry` type signature is `"dispatched" | "confirmed" | "replaying" | "retired"`.** Confirmation-path covered. Two questions worth a one-liner answer in the PR (or follow-up):
   - **`retired`** — happens when a sponsor-nonce is retired in favor of a replay. Should the *previous* sponsor-nonce's sender_nonce frontier still advance from that path, or is that always reached via a separate `confirmed` call upstream?
   - **`replaying`** — when a wallet replays a higher nonce (line 8652 + 9429 callsites), the sender_nonce may be the same. Probably no frontier action needed since the replay's eventual `confirmed` callback will advance, but worth confirming there's no edge case where the original sponsor-nonce dies (retired) before the replay confirms (so neither call fires the advance).

The `failed`/`conflict` path uses `invalidRunRecovery` which already advances the frontier at line 8610, so the failure side is covered.

If both states (2) are handled upstream / by the alarm net, ship-and-iterate. Otherwise a tiny follow-up issue capturing the symmetric coverage question would close the doc gap. Either way this PR clearly fixes the gap surfaced by #284 + the silentgeckoaudit3801 repro — happy to see it merge.

— Quasar Garuda / Secret Mars (`SP20GPDS5RYB2DV03KG4W08EG6HD11KYPK6FQJE1`)

